PR #33, opened 2018-10-17
Add auction list

### LINKED ISSUE #2: Get a list of domains to auction
Allow:
- [ ] Secure communication with the Registry (HTTPS, token or other type of security)
- [ ] Connect to Registry via RESTful API, and get a list of domains that are up for auction in the near future.

Actual implementation details to follow once the Registry API is decided.

maciej-szlosarczyk:
Fixes #2, it's the root page.

Also adds a form to create auctions manually. We won't use it, but because of the open-source nature, it should be there. Currently, the administrator can create an auction with only three parameters:

* `domain_name`, not validated at all, must not be empty.
* `starts_at`, some time, cannot be in the past
* `ends_at`, some time, cannot be earlier than `starts_at` (validation, db constraint)
* Auction must be unique for its duration for the domain_name (constraint). There cannot be 2 auctions for the same domain happening at the same time

An administrator can create new auctions and delete existing ones if they have not started. If the auction is in progress or already finished, there is no UI to modify it.


maciej-szlosarczyk:
This can probably be merged even if there are errors in it (should have their own tickets).

vohmar:
merge conflict with master after merging remove-vat-code-unique-constraint

maciej-szlosarczyk:
Fixed now.

PR #34, opened 2018-10-17
Change vat code requirements

### LINKED ISSUE #26: VAT code in billing profile has unique constraint
System does not allow adding multiple billing profiles with the same vat code in the system. While this is OK for a single user it is not for the whole system as every organisatsion can have multiple representatives. 

At the same time adding a billing profile must not give access to any of the invoices assigned to that company (bids are done by a unique users and only that user has access to the billing and authorisation code related to the winning bid)

Also error handlung is missing for adding multiple profiles with the same vat code for the same user - error 500 is returned.

maciej-szlosarczyk:
* It can be blank for businesses
* It can repeat between users
* It has to be unique for a specific user

Fixes #26

### DIFF
--- a/app/controllers/billing_profiles_controller.rb
+++ b/app/controllers/billing_profiles_controller.rb
@@ -5,6 +5,7 @@ class BillingProfilesController < ApplicationController
   # GET /billing_profiles
   def index
     @billing_profiles = BillingProfile.accessible_by(current_ability)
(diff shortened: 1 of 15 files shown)


maciej-szlosarczyk:
Removed the obsolete legal_entity checkbox. Also fixed #35 and an information leak where administrator could see all the billing profiles in the normal user interface.

PR #50, opened 2018-11-15
Create results after auction is finished

### LINKED ISSUE #8: Async job processing with delayed_job
Use delayed_job for processing tasks that require background processing:

- Sending emails to winners of auctions
- Fetching the list of domains to auction
- Sending auction results to registry

---
The reason why we should go with `delayed_job`, and not `sidekiq` is that the latter does only support redis as backend, thus adding one more infrastructure element that needs to be managed and updated.

maciej-szlosarczyk:
Fixes #8. Results are calculated automatically when an administrator opens the auction page or any user opens the offers page (this will change in the future to be more reliable.

When an auction is finished, the user gets an email informing him/her of that fact. If there is none, no email is sent. Later on, that link will contain a form for the user to set up the invoice and pay it.

Adds `payment_term` Setting, which allows the administrator to setup how many days from now should be the invoice be valid for. Currently only used in the email.


maciej-szlosarczyk:
Need some additional changes in staging infrastructure.

maciej-szlosarczyk:
Hey @vohmar, it is a bit complicated to deploy this to staging without merging it to master first, since it contains a breaking infrastructure change. I'll merge it and then make required adjustments in staging.

PR #130, opened 2019-03-22
Update production.rb

teadur:


teadur:
Default logger to syslog

PR #149, opened 2019-03-26
Fix BillingProfile#edit error

maciej-szlosarczyk:
When a value is not given, it should not be a 0-character string, but
a nil. This caused an error in production.

https://errbit.internet.ee/apps/5bd9d0abd9d1af253512a1df/problems/5c995ba1d9d1af25351307e4


maciej-szlosarczyk:
Additional fixes:

- Fix `et.devise.sessions.signed_in`
- Fix terms and conditions text on TARA form.

PR #159, opened 2019-03-27
Fix everypay error

maciej-szlosarczyk:


maciej-szlosarczyk:
@vohmar This is ready, can be merged and deployed.

PR #161, opened 2019-03-27
Do not allow to create duplicate offers

maciej-szlosarczyk:
Steps to reproduce unwanted behaviour:

1. Open new offer page twice for the same user in two browser windows/tabs.
2. Try to submit two offers.

The later one should redirect you to view of existing offer. This is not a fix for someone submitting offers exactly at the same time, that would be an unique index. We might need an ability to have multiple offers from the same user from different auction types.

FYI @teadur


vohmar:
translation for the error message missing. Please use "Oled sellele oksjonile pakkumuse juba esitanud."

image

PR #163, opened 2019-03-28
Add paginate block to invoices

maciej-szlosarczyk:


maciej-szlosarczyk:
This can be merged without testing, we use the same functionality in many other places.

PR #180, opened 2019-03-29
Handle VAT in Estonia according to business rules

### LINKED ISSUE #179: VAT is not added for Estonian VAT liable billing profiles
If user creates an invoice for a company that is VAT liable the VAT is not added to the invoice even if the company is Estonian.

* VAT must be always added if the billing profile is Estonian
* VAT must never be added if the billing profile is outside of EU
* VAT must be added if the billing profile is from EU but is without VAT number (vat of that country must be added in this case)
* VAT must not be added if the billing profile is from EU (but not from Estonia) and has VAT number

maciej-szlosarczyk:
* For Estonian person, VAT should be applied always
* For EU private person, VAT should be applied
* For EU person with a VAT number, VAT should not be
  applied (chargeback)
* For non-EU person, do not calculate VAT

Fixes #179


maciej-szlosarczyk:
@vohmar You can test this and if all is correct, deploy to production. It should change VAT numbers displayed on all existing invoices, even the paid ones as the value is dynamically assigned.

vohmar:
for eu org without vat number it still shows 0% vat while their local vat must be applied in this case. 

for any person (legal or private) that is registered in EU and do not have vat number (not vat liable) should have their local vat applied to the invoices

maciej-szlosarczyk:
There was a bug related to this 3 days ago. It is already fixed in #149 

There was latent data on staging that causes this issue. I wrote a migration that changes the incorrect data, it technically is a production data migration, but does nothing destructive. Changes billing profiles with vat_code `""` to `NULL` to enable correct VAT calculation.

You can retest.

PR #206, opened 2019-04-04
Remove dependencies we do not use

### LINKED ISSUE #68: Remove unecessary Semantic UI modules
Once the UI has been stabilized, remove unused Javascript modules from app/javascript/src/semantic-ui. We currently use only a few and there is no need to have them all in the repository.

### LINKED ISSUE #157: Remove whenever gem
It's usefulness is limited, as we can write our own cron scripts against rails runner.

maciej-szlosarczyk:
Removes semantic-ui modules. Those can still be found at in our other github repository at:
https://github.com/internetee/auction_css

Fixes #68 

Removes whenever gem (we use ansible for those). Fixes #157


maciej-szlosarczyk:
There is not much to test here, most important part is that dropdowns, checkboxes and mobile menu continue to work as they do now.

PR #211, opened 2019-04-05
Change payment_term setting description

### LINKED ISSUE #208: Valid period for invoices 1 day too long
The validity period for for the invoices should be 7 days. As invoice creation job is run at the beginning of a day the issue date is also counted in as a full day. Because of that traditional calculation issue_date + 7 results in a payment period of full 8 days.

To fix the calculation  one option would be to add a setting for payment term - last day included/excluded. Meaning that if period of 7 is specified, does the calculated due date mean that user has to pay the invoice before that date or at that date the latest. On the invoice we always show date as inclusive. That means if  setup is last day excluded the due date on invoice needs to be -1 from actual calculated due date.

Other approaches to solve this issue are possible.

maciej-szlosarczyk:
Fixes #208


vohmar:
merge conflict

maciej-szlosarczyk:
Fixed, mr. @vohmar

PR #213, opened 2019-04-05
Add foreign key constraint on invoice

maciej-szlosarczyk:
Companion to #209


maciej-szlosarczyk:
Does not need to be tested, but will create a conflict with #211 because both run a migration.

PR #224, opened 2019-04-15
Load Raleway font from NPM package

### LINKED ISSUE #223: Ensure that Raleway font is always used
In production, we do not use the right font. We should be always using Raleway, which a part of our visual identification, but instead, we use system default sans-serif.

The problem is related to building JS assets.

maciej-szlosarczyk:
Fixes #223


maciej-szlosarczyk:
Example of how it is:

Screenshot 2019-04-15 at 09 47 17

Example of hos it should be:

Screenshot 2019-04-15 at 09 47 07

PR #243, opened 2019-04-25
Fix banklink error

maciej-szlosarczyk:
Fixes error we are seeing in Errbit. Please test that payment methods continue to work.


maciej-szlosarczyk:
https://errbit.internet.ee/apps/5bd80564d9d1af253512a1c3/problems/5cc16e21d9d1af1242c22951

Example issue.

PR #258, opened 2019-05-03
Add information about background jobs to readme

maciej-szlosarczyk:


maciej-szlosarczyk:
Also disable a special partition used for cache by Chromium, which only affects test:

```
// The /dev/shm partition is too small in certain VM environments, causing
// Chrome to fail or crash (see http://crbug.com/715363). Use this flag to
// work-around this issue (a temporary directory will always be used to create
// anonymous shared memory files).
```

PR #259, opened 2019-05-03
Disable country dropdown

### LINKED ISSUE #256: Disable country code update for e-authenticated users
for electronically authenticated users we fill the ident code and country code automatically, but in the user profile the country code is editable, ident code is not. As the primary use of country code is to be able to validate and verify the identity of the user using their identity code, changing it will make the identity code together with the updated country code invalid.

user should not be able to change the country code if it is automatically set using valid electronic id.

maciej-szlosarczyk:
Fixes #256

### DIFF
--- a/app/views/auth/tara/_form.html.erb
+++ b/app/views/auth/tara/_form.html.erb
@@ -12,15 +12,10 @@
                 </div>
             </div>
             <div class="field">
-                <%= f.label :country, t('users.country') %>
+                <%= f.label :country, t('users.country_code') %>
                 <div class="ui disabled input">
-                    <%= f.select :country_code,
-                    options_for_select(
-                        Countries.for_selection,
-                        user.country_code || Setting.default_country
-                    ),
-                    { readonly: true },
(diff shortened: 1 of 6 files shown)


maciej-szlosarczyk:
Also disable a special partition used for cache by Chromium, which only affects test:

```
// The /dev/shm partition is too small in certain VM environments, causing
// Chrome to fail or crash (see http://crbug.com/715363). Use this flag to
// work-around this issue (a temporary directory will always be used to create
// anonymous shared memory files).
```

vohmar:
does not seem to have an effect - the dropdown is still there and user is able to set and save new value

image

also I just spotted few translation issue in that view - column headers. Hope it is ok to fix these here without separate ticket.

Data from identity document = Andmed elektroonselt isikutunnistuselt
Contact data = Kontaktandmed

maciej-szlosarczyk:
@vohmar now should be better. It did work only in one browser, it seems. I had to replace the dropdown with a text field, so now an ISO-2 country code is displayed, but it is not editable.

PR #275, opened 2019-06-04
Update non-critical javascript dependencies

maciej-szlosarczyk:


maciej-szlosarczyk:
Will probably create a conflict with #277, but this can be safely merged ahead without testing.

PR #280, opened 2019-06-06
Fix omniauth CVE

maciej-szlosarczyk:
Protect against CVE-2015-9284

Use POST method to initialize Tara request. No further action should be required anymore.

Please ensure that Tara login and user creation continues to work after this.


vohmar:
tara auth and user creation continues to work as expected.

maciej-szlosarczyk:
Additional comment here:

We were not vulnerable to this exploit because Tara does not allow a user to be persistently logged in But, if we were to ever introduce Google's or Facebook's SSO, then we would be exposed.

Issue #128, opened 2019-03-22
Remove N+1 from auctions index


maciej-szlosarczyk:
Auction index creates a lot of queries for each user to find if the user already has an auction for specific domain (participant) or check what is current highest offer (administrator). Those should be removed, and that data should be retrieved with one SQL.

maciej-szlosarczyk:
Fixed in #299

PR #287, opened 2019-06-13
Orderable tables

### LINKED ISSUE #173: Admin: sorting
Add sorting fuctionality to the reports (asc and descending) - auctions, results, billing profiles, users, bans

maciej-szlosarczyk:
Adds ordering for administrators as well as ordering of auctions by domain name and finishing time for participant users.

An administrator can order:
- auctions
- bans
- results
- users
- invoices

Ordering can also be applied to search results. Ordering is chainable, so you can order by N columns.

Fixes #173


vohmar:
* please add the sorting options to billing profiles as well (user and nama columns at least)
* this is nice to have but it would be interesting to have sorting option on auctions highest price and offers columns as well to easily get access to some interesting stats

maciej-szlosarczyk:
Billing profiles are now also included, you can order them by user surname, profile name and profile VAT code.

Sorting on auction highest price is tricky until #128 is resolved. Those fields are not backed by easily accessible database columns. When we/I solve #128, I'll add ordering by highest price and number of offers per auction.

PR #291, opened 2019-06-14
add view port for mobile devices

rieussec:
The site is very well thought for the responsive mobile car except that it misses the header, I propose to add it!


maciej-szlosarczyk:
Good catch there! Nothing really to test, can be merged right away.

PR #298, opened 2019-07-05
Update test dependencies

maciej-szlosarczyk:
Fixes security alert discovered by npm-security:

https://github.com/wycats/handlebars.js/commit/edc6220d51139b32c28e51641fadad59a543ae57

Also fixes a deprecation warning coming from `webdrivers` gem:

```
nil versions are discouraged and will be deprecated in Rubygems 4
WARN Webdrivers Driver caching is turned off in this version, 
     but will be enabled by default in 4.x. 
     Set the value with `Webdrivers#cache_time=` in seconds
```

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -68,13 +68,13 @@ GEM
     builder (3.2.3)
     byebug (11.0.1)
     cancancan (2.3.0)
-    capybara (3.18.0)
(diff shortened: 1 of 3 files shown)


maciej-szlosarczyk:
This has no impact on production code, can be merged right away.

PR #299, opened 2019-07-08
Add sorting by highest price and number of offers

maciej-szlosarczyk:
In administrator UI, you can now sort auctions by highest offer and number of offers.A participant can also order auctions by a new parameter - his/her own biggest offer. Both also work in search.

Fxes #128, should improve overall performance.

Things to check while testing:
- JSON API continues to return all active auctions, even if there's more than 20 (that's per page limit)
- No information leaks from auction index page. Particularly, no user can get different's user offers.


app/models/admin_auction_decorator.rb
```diff
@@ -0,0 +1,77 @@
+# Query decorator for Auctions. Implements Orderable interaface
```
artur-beljajev: interaface = interface

Issue #335, opened 2019-08-25
Upgrade to Rails 6


artur-beljajev:
- [x] Update `delayed_job` and `delayed_job_active_record` gems
- [x] Update `jbuilder` gem
- [x] Update `devise` gem

PR #336, opened 2019-08-25
Update `delayed_job` gem

artur-beljajev:
#335


artur-beljajev:
See https://github.com/internetee/auction_center/pull/337

PR #357, opened 2019-10-08
Auctions: Make domain table stackable for mobile users

### LINKED ISSUE #354: Mobile view is unusable
Mobile view is unusable especially in portrait mode.
image

it is important to fit the whole domainname in there. One option is to remove all the other columns in mobile view and leave only the domainname. User can tap on the name to see details of the auction, including their current bid, time of auction end and can make a bid or edit current one

karlerikounapuu:
Makes the domain table stackable for mobile users on `/auctions` and `/auctions/search`. 

image

Closes #354.

### DIFF
--- a/app/javascript/src/global/_common.scss
+++ b/app/javascript/src/global/_common.scss
@@ -158,4 +158,22 @@ strong {
     @include max-screen(991px) {
         margin: 40px 0;
     }
+
+    .auctions-table-mobile-infotainment {
+        @media only screen and (min-width: 768px) {
+            display: none;
+        }
+    }
+
+    #auctions-table > thead > tr {
+        th.auctions-table-cell-name-no-mobile {
+            @media only screen and (max-width: 767px) {
+                display: none !important;
+            }
+        }   
+    }
 }
+
+.ui.table.stackable thead {
+    background: #212224;
+  }
--- a/app/javascript/src/semantic-ui/definitions/collections/grid.less
+++ b/app/javascript/src/semantic-ui/definitions/collections/grid.less
@@ -1821,9 +1821,6 @@
   .ui.grid.grid.grid > .row > [class*="widescreen only"].column:not(.mobile) {
     display: none !important;
   }
-  .auction[class*="mobile only"]:not(.auction[class*="tablet computer large screen widescreen"]) {
-    display: none !important;
-  }
 }
 /* Tablet Only Hide */
 @media only screen and (min-width: @tabletBreakpoint) and (max-width: @largestTabletScreen) {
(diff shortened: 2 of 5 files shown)


app/views/auctions/search.html.erb
```diff
@@ -26,8 +26,20 @@
     <div class="margin-block narrow">
         <div class="grid one column">
             <div class="column">
-                <table class="ui table selectable unstackable fixed" id="auctions-table">
-                    <thead>
+                <table class="ui table selectable mobile stackable fixed" id="auctions-table">
+                    <thead class="mobile-show">
```
artur-beljajev: Can we use semantic class name and move logic (I assume it is visible only in mobile, based on the name) to CSS?

karlerikounapuu:
> 1. Any chance to avoid using `!important`?

Possible to avoid `!important` tag in mobile-show class. Removed with last commit. Not that possible with other one since Semantic UI is pretty aggressive when it comes to pushing grid rules.

> 2. Why `scope="col"` is needed?

Came from habit, to be honest. Removed with last commit.

app/views/auctions/search.html.erb
karlerikounapuu: Renamed to "mobile show", semantic enough I believe. Wouldn't really define deeper selector scope, because it could be heavily referenced from other views as well.

artur-beljajev:
@karlerikounapuu I wonder if app-specific overrides should really reside in `app/javascript/src/semantic-ui/`. I can't see many changes here https://github.com/internetee/auction_center/commits/master/app/javascript/src/semantic-ui.

Perhaps we should put it here `app/javascript/src/common` or here `app/javascript/src/global/`?

karlerikounapuu:
> @karlerikounapuu I wonder if app-specific overrides should really reside in `app/javascript/src/semantic-ui/`. I can't see many changes here https://github.com/internetee/auction_center/commits/master/app/javascript/src/semantic-ui.
> 
> Perhaps we should put it here `app/javascript/src/common` or here `app/javascript/src/global/`?
> 
> P.S. We outsource front-end part. It should be easy to integrate it.

@artur-beljajev Changing file path for overrides makes sense. I've moved them to `app/javascript/src/global/`

app/javascript/src/global/_common.scss
```diff
@@ -158,4 +158,22 @@ strong {
...
         margin: 40px 0;
     }
+
+    .auctions-table-mobile-infotainment {
+        @media only screen and (min-width: 768px) {
+            display: none;
+        }
+    }
+
+    #auctions-table > thead > tr {
```
artur-beljajev: Such nesting is quite brittle. Might be better to use classes.

app/javascript/src/semantic-ui/definitions/collections/grid.less
```diff
@@ -1917,4 +1917,4 @@
 }
 
 
-.loadUIOverrides();
+.loadUIOverrides();
```
artur-beljajev: What is changed here?

app/javascript/src/semantic-ui/definitions/collections/table.less
```diff
@@ -1144,4 +1144,4 @@
   font-size: @large;
 }
 
-.loadUIOverrides();
+.loadUIOverrides();
```
artur-beljajev: What is changed here?

PR #364, opened 2019-10-15
Auctions: Add option to order auction list on email

### LINKED ISSUE #355: Add option to order auction list on email
A user has requested an option to get the auction list on their email every day.

This should be optional, disabled by default.

Maybe add this option to wishlist view and reference it from domain list as well

karlerikounapuu:
Closes #355 

Adds an "email" icon to both `auctions/` and `auctions/search` pages. Redirects user to either `new_user_path` or `user_path`, which exposes a function for client to opt-in / opt-out from auction newsletters.


app/javascript/src/global/_common.scss
```diff
@@ -170,10 +170,20 @@ strong {
             @media only screen and (max-width: 767px) {
                 display: none !important;
             }
-        }   
+        }
     }
 }
 
+.auction-newsletter-link {
+    margin-left: 0.75em !important;
```
artur-beljajev: I understand it might be required by the current framework, but I still wonder if there is any chance to avoid `!important` (way too many of them :( ). It will be a nightmare to maintain such CSS later.

app/views/admin/users/_form.html.erb
```diff
@@ -48,7 +48,13 @@
                     <% end %>
                 </div>
             </div>
-
+            <h3 class="p3"><%= t('users.form.newsletters') %></h3>
```
artur-beljajev: How about using `t('.daily_summary')` shortcut?

app/views/admin/users/_form.html.erb
```diff
@@ -48,7 +48,13 @@
...
+            <div class="field">
+                <div class="ui checkbox">
+                    <%= f.check_box :auction_newsletter, { value: user.auction_newsletter }, "true", "false" %>
```
artur-beljajev: Value is taken from model automatically. There is no need to explicitly provide it.

artur-beljajev: What these `true` and `false` mean?

app/mailers/notification_mailer.rb
```diff
@@ -11,6 +11,12 @@ def daily_summary_email(addressee, summary_report = nil)
     mail(to: addressee.email, subject: t('.subject', date: Date.yesterday))
   end
 
+  def daily_auctions_broadcast_email(addressee, auctions)
```
artur-beljajev: Perhaps `recipient` is more readable?

app/mailers/notification_mailer.rb
```diff
@@ -11,6 +11,12 @@ def daily_summary_email(addressee, summary_report = nil)
...
+    I18n.locale = addressee.locale
+    @auctions = auctions
+    mail(to: addressee.email, subject: t('.subject'))
```
artur-beljajev: Subject is i18n-enabled by default. How about removing it and rely on automatic look-up?

app/jobs/daily_broadcast_auctions_job.rb
```diff
@@ -0,0 +1,16 @@
+class DailyBroadcastAuctionsJob < ApplicationJob
+  def perform
+    User.where(auction_newsletter: true).each do |subscriber|
```
artur-beljajev: Probably `subscriber` -> `user`. I can't find such entity as subscriber.

app/jobs/daily_broadcast_auctions_job.rb
```diff
@@ -0,0 +1,16 @@
...
+      NotificationMailer.daily_auctions_broadcast_email(subscriber, auctions_today).deliver_later
+    end
+  end
+
+  def auctions_today
+    timeframe = Time.zone.today.beginning_of_day..Time.zone.today.end_of_day
+    Auction.where(starts_at: timeframe).to_ary
```
artur-beljajev: Why `.to_ary` is needed?

app/mailers/notification_mailer.rb
artur-beljajev: How about using keyword arguments?

test/jobs/daily_broadcast_auctions_job_test.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def test_sends_email_to_subscribers
+    perform_enqueued_jobs do
+      DailyBroadcastAuctionsJob.perform_now
+    end
+
+    assert_not(ActionMailer::Base.deliveries.empty?)
+    email = ActionMailer::Base.deliveries.last
+    assert_equal(['user@auction.test'], email.to)
```
artur-beljajev: This should be tested in mailer test.

app/javascript/src/global/_common.scss
karlerikounapuu: I agree. Eliminated one !important tag. In overall, I don't see any other reasonable methods on how to overwrite these values because fair amount of rules are forced like this as well. But as long as overrides are in single file(s) systematically, I'm not concerned.

app/views/admin/users/_form.html.erb
karlerikounapuu: Mhm, both removed. By syntax they're meant to define output of checkbox (1 or 0, true or false etc), but not really needed in this scenario. As other checkboxes (tos, for example) were implemented same way, I decided to follow this pattern.

app/jobs/daily_broadcast_auctions_job.rb
karlerikounapuu: Not sure what I'm missing, but without `.to_ary` I receive `ActiveJob::SerializationError ` when creating domain list in mail view. Tips?

yulgolem: I suppose its totally OK. The result of #auctions_today method is used in `NotificationMailer.daily_auctions_broadcast_email(subscriber, auctions_today).deliver_later`
The point is in `deliver_later`, which is processed by ActiveJob.
ActiveJob currently supports String, Integer, Float, NilClass, TrueClass, FalseClass, BigDecimal, Symbol, Date, Time, DateTime, ActiveSupport::TimeWithZone, ActiveSupport::Duration, Hash, ActiveSupport::HashWithIndifferentAccess, Array or GlobalID::Identification instances (https://api.rubyonrails.org/classes/ActiveJob/SerializationError.html) , so here we can:
- either pass an array of something and find object in mailer;
- or use custom serializer which can serialize ActiveRecord_Relation.

app/mailers/notification_mailer.rb
karlerikounapuu: What's the benefit of keyword arguments in this scenario?

app/jobs/daily_broadcast_auctions_job.rb
```diff
@@ -0,0 +1,19 @@
+class DailyBroadcastAuctionsJob < ApplicationJob
+  def perform
+    User.where(daily_summary: true).each do |user|
```
artur-beljajev: The knowledge that a `User` model has `daily_summary` model belongs to a model. I realize that `ActiveRecord` pattern (unfortunately) encourages to treat a model more like a data bag instead of real object, but perhaps it still would be better to keep that knowledge in the model, when possible?

Something like `User.subscribed_to_daily_summary`.

app/javascript/src/global/_common.scss
```diff
@@ -181,3 +181,13 @@ strong {
 .auction-domain-name {
     word-break: break-all;
 }
+
+.daily-summary-link {
+    margin-left: 0.75em !important;
```
artur-beljajev: No way to avoid `important`? (sorry for repeating it on and on :) ).

app/jobs/daily_broadcast_auctions_job.rb
```diff
@@ -0,0 +1,19 @@
...
+      NotificationMailer.daily_auctions_broadcast_email(
+        recipient: user,
+        auctions: auctions_today
+      ).deliver_later
+    end
+  end
+
+  def auctions_today
+    timeframe = Time.zone.today.beginning_of_day..Time.zone.today.end_of_day
+    Auction.where(starts_at: timeframe).to_a
```
artur-beljajev: Same as above: how about something like `Auction.started_within(period)`?

app/mailers/notification_mailer.rb
```diff
@@ -11,6 +11,11 @@ def daily_summary_email(addressee, summary_report = nil)
     mail(to: addressee.email, subject: t('.subject', date: Date.yesterday))
   end
 
+  def daily_auctions_broadcast_email(recipient:, auctions:)
+    @auctions = auctions
+    mail(to: recipient.email, subject: t('.subject'))
```
artur-beljajev: Subject is localized by default. Better to remove it, and instead use translation files when needed.

app/mailers/notification_mailer.rb
```diff
@@ -11,6 +11,11 @@ def daily_summary_email(addressee, summary_report = nil)
     mail(to: addressee.email, subject: t('.subject', date: Date.yesterday))
   end
 
+  def daily_auctions_broadcast_email(recipient:, auctions:)
```
artur-beljajev: User is not a recipient, still. One may think there is a `Recipient` role (=class), but here is none. By my previous suggestion I meant passing user's email as a `recipient` value. Is it really needed to pass the whole object here?

app/views/admin/users/_form.html.erb
```diff
@@ -48,7 +48,13 @@
                     <% end %>
                 </div>
             </div>
-
+            <h3 class="p3"><%= t('users.daily_summary') %></h3>
```
artur-beljajev: Using short form `t('.')` enables deeper namespacing (which is a good thing).

app/views/admin/users/_form.html.erb
```diff
@@ -48,7 +48,13 @@
...
+            <div class="field">
+                <div class="ui checkbox">
+                    <%= f.check_box :daily_summary %>
+                    <%= f.label :daily_summary, t('users.form.daily_summary_intention') %>
```
artur-beljajev: Same here.

app/views/admin/users/show.html.erb
```diff
@@ -60,6 +60,10 @@
                             <td><strong><%= t('created_at') %></strong></td>
                             <td><%= I18n.l(@user.created_at) %></td>
                         </tr>
+                        <tr>
+                            <td><strong><%= t('users.daily_summary') %></strong></td>
+                            <td><%= @user.daily_summary? ? I18n.t('yes') : I18n.t('no')%></td>
```
artur-beljajev: I would use `t` here, since `I18n.t` is not the same as `t`. 
For example, `config.action_view.raise_on_missing_translations` has no effect when using `I18n.t` (which is bad).

app/views/auctions/index.html.erb
```diff
@@ -19,6 +19,15 @@
                 <% end %>
             <% end %>
         </div>
+        <% if current_user %>
```
artur-beljajev: Indent it for better readability.

app/views/auctions/search.html.erb
```diff
@@ -19,6 +19,15 @@
                 <% end %>
             <% end %>
         </div>
+        <% if current_user %>
```
artur-beljajev: Same here.

app/views/notification_mailer/daily_auctions_broadcast_email.html.erb
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta content='text/html; charset=UTF-8' http-equiv='Content-Type'/>
+    </head>
+    <body>
```
artur-beljajev: It's better to move this to layout instead.

app/views/users/_form.html.erb
```diff
@@ -49,6 +49,13 @@
                     <% end %>
                 </div>
             </div>
+            <h3 class="p3"><%= t('users.daily_summary') %></h3>
```
artur-beljajev: Why `h3`? It is really a header?

app/views/users/_form.html.erb
```diff
@@ -49,6 +49,13 @@
...
+            <div class="field">
+                <div class="ui checkbox">
+                    <%= f.check_box :daily_summary %>
+                    <%= f.label :daily_summary, t('.daily_summary_intention') %>
```
artur-beljajev: Aren't labels translated automatically? Why `t('.daily_summary_intention')` is needed here and why `intention` word is needed here?

app/views/users/show.html.erb
```diff
@@ -52,6 +52,10 @@
                             <td><strong><%= t('users.terms_and_conditions') %></strong></td>
                             <td><%= link_to t('.review_terms_and_conditions'), Setting.terms_and_conditions_link %></td>
                         </tr>
+                        <tr>
+                            <td><strong><%= t('users.daily_summary') %></strong></td>
+                            <td><%= @user.daily_summary? ? I18n.t('yes') : I18n.t('no')%></td>
```
artur-beljajev: Check above comments. Also avoid cluttering global namespace with such values as `yes` and `no`.

config/locales/notification_mailer.en.yml
```diff
@@ -9,3 +9,7 @@ en:
 
       registration_due_dates: "Domains due to be registered tomorrow:"
       bans: "Bans issued yesterday:"
+    daily_auctions_broadcast_email:
+      subject: "Today's .EE domain auction list is available!"
```
artur-beljajev: Use default translation path (example: https://github.com/internetee/registry/blob/master/config/locales/mailers/contact.en.yml). See comments above for details.

test/jobs/daily_broadcast_auctions_job_test.rb
```diff
@@ -0,0 +1,31 @@
+require 'test_helper'
+
+class DailyBroadcastAuctionsJobTest < ActiveJob::TestCase
+  include ActionMailer::TestHelper
+  def setup
```
artur-beljajev: Use callback instead (preferred) or call `super`, otherwise parent method is not called.

test/mailers/notification_mailer_test.rb
```diff
@@ -27,4 +30,16 @@ def test_mail_gets_delivered
...
   end
+
+  def test_sends_auctioned_domain_list_to_subscribers
+    perform_enqueued_jobs do
+      DailyBroadcastAuctionsJob.perform_now
+    end
+
+    assert_not(ActionMailer::Base.deliveries.empty?)
+    email = ActionMailer::Base.deliveries.last
+    assert_equal(['user@auction.test'], email.to)
```
artur-beljajev: It's not clear where does this value come from. Can we explain it (with code) in the "given" phase?

test/system/admin/admin_jobs_test.rb
```diff
@@ -21,8 +21,9 @@ def test_administrator_can_schedule_jobs_on_demand
     assert_text('InvoiceCreationJob')
     assert_text('InvoiceCancellationJob')
     assert_text('DailySummaryJob')
+    assert_text('DailyBroadcastAuctionsJob')
 
-    assert(page.has_button?('Run', count: 4))
+    assert(page.has_button?('Run', count: 5))
```
artur-beljajev: Seems like magic number. What is it responsible for? Can we avoid it?

app/views/admin/users/_form.html.erb
karlerikounapuu: Scope is `users.form` and `daily_summary` is already defined in `users` scope (same translation is used on profile page). Not going to change this one because in my opinion it's not good idea to duplicate translations, is it?

artur-beljajev: It might make more sense to use `admin` namespace, since this is what it represents.

app/views/users/_form.html.erb
karlerikounapuu: `daily_summary_intention` provides reason / intention for the user why this checkbox should be checked. Just like ToS.

artur-beljajev: But why not just `daily_summary`? What happens if you remove `t` call completely? They should be translated automatically, as far as I know.

app/views/admin/users/_form.html.erb
artur-beljajev: > In case if the key is too generic hard to track the relationship between the source and actual place of display on the page

This is what I meant. Taken from first link in Google: https://dev.to/omaiboroda/three-ways-to-name-i18n-translation-keys-2fed

app/mailers/notification_mailer.rb
karlerikounapuu: How does mailer know which locale to use if it's not aware of user and it's locale?

app/views/admin/users/_form.html.erb
artur-beljajev: Rails translations seem have nothing to do with Gettext, where such practice is used (one single translation "key").

app/mailers/notification_mailer.rb
karlerikounapuu: Btw thanks for bringing this up, because just discovered it doesn't send email in et locale anyways at the moment :)

app/views/users/_form.html.erb
karlerikounapuu: Sure, it would auto-translate to "Daily summary to email" and "Igapäevane oksjoninimekiri meilile", but I had UX in mind with this one. Seemed nice touch to provide some additional, well-explained comment about this checkbox, but guess it's enough now. Removing it.

app/views/users/show.html.erb
karlerikounapuu: These values were already present in `et.yml` beforehand, so thought I would simply use them to provide better context. It would provide "true" and "false" to et users without it, which seems like "translation error" to me.

config/locales/notification_mailer.en.yml
karlerikounapuu: I agree. But all mailers in `auction_center` app are using same (not perfect) path so I simply followed the pattern. I'm opening a new ticket because moving them all right now is not the scope of this PR.

app/views/admin/users/_form.html.erb
karlerikounapuu: Alright, makes sense. Making changes.

test/jobs/daily_broadcast_auctions_job_test.rb
karlerikounapuu: `super` is already present in setup method, am I missing something? What's the benefit of preferring callback?

app/views/users/_form.html.erb
karlerikounapuu: Yes, it was header, indeed. But as I removed `daily_summary_intention`, it's no longer needed.

config/locales/notification_mailer.en.yml
artur-beljajev: Perhaps we could still start using default paths at least for the things directly connected to this PR?

app/mailers/notification_mailer.rb
artur-beljajev: > How does mailer know which locale to use if it's not aware of user and it's locale?

You need just user's locale, not `I18n.locale` right? Then one option is pass the whole object `user` (`def daily_auctions_broadcast_email(user:)`, like you did initially, just with the different param name).

P.S.

> is it really needed to pass the whole object here?

I just asked whether is it needed to pass it, not that you did it wrong :)

artur-beljajev: P.S. II: example of using "built-in" translation if you need to pass some data to translation value https://github.com/internetee/registry/blob/master/app/mailers/invoice_mailer.rb#L5. Not sure if you need it in this case, but this is a good way to use standard translation keys.

test/mailers/notification_mailer_test.rb
```diff
@@ -27,4 +30,17 @@ def test_mail_gets_delivered
     assert_equal(['administrator@auction.test'], email.to)
     assert_equal('Daily summary for 2010-07-04', email.subject)
   end
+
+  def test_sends_auctioned_domain_list_to_subscribers
+    perform_enqueued_jobs do
+      DailyBroadcastAuctionsJob.perform_now
```
artur-beljajev: Here, a job is tested in mailer test. Might be something like
```
email = ContactMailer.email_changed(contact: @contact, old_email: 'john-old@inbox.test')
              .deliver_now
```

More examples: https://github.com/internetee/registry/tree/master/test/mailers

test/mailers/notification_mailer_test.rb
```diff
@@ -27,4 +30,17 @@ def test_mail_gets_delivered
...
+    end
+
+    assert_not(ActionMailer::Base.deliveries.empty?)
+    email = ActionMailer::Base.deliveries.last
+    # User email, which is provided by DailyBroadcastAuctionsJob
```
artur-beljajev: This comment can be removed if we would not be testing a job in mailer test.

test/mailers/notification_mailer_test.rb
```diff
@@ -27,4 +30,17 @@ def test_mail_gets_delivered
...
+    assert_equal(['user@auction.test'], email.to)
+    assert_equal("Today's .EE domain auction list is available!", email.subject)
```
artur-beljajev: Are double quotes really needed here?

test/system/admin/admin_jobs_test.rb
artur-beljajev: It has some unrelated reasons to fail, therefore it's quite brittle: adding a new job immediately makes it failing. 
Based on its name `administrator_can_schedule_jobs_on_demand`, it should fail if an admin cannot schedule jobs anymore, which is a false statement if I just add a new job.

So something is wrong here. Perhaps as a first step, `ApplicationJob.descendants.count` might be a more readable alternative for that magic value...

It's indeed seems like a topic of a new PR.

artur-beljajev:
@karlerikounapuu There are still unrelated changes in `db/structure.sql`, therefore I cannot review it.

test/mailers/notification_mailer_test.rb
karlerikounapuu: Yes they were, because mail title itself contained quote

karlerikounapuu:
> * There are still unrelated changes in `db/structure.sql`, therefore I cannot review it.
> * Also, please check double quotes usage. There are some places where I don't see a point using them. Consider using single ones when no interpolation is needed.

@artur-beljajev 
Fixed `db/structure.sql`. Turned out it's generated using postgres@9.4 binary, whereas I've been using 9.6 because it was references in travis conf. Lesson learned 😀

Also removed unnessecary quotation marks.

app/models/auction.rb
```diff
@@ -107,4 +107,8 @@ def in_progress?
       false
     end
   end
+
+  def self.started_within(period)
```
yulgolem: Did you consider to move it to scope?

app/models/user.rb
```diff
@@ -130,4 +130,8 @@ def self.from_omniauth(omniauth_hash)
       end
     end
   end
+
+  def self.subscribed_to_daily_summary
```
yulgolem: Same here - maybe move it to scope? Just asking.

db/migrate/20191008124157_add_daily_summary_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddDailySummaryToUser < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :daily_summary, :boolean, default: false
```
yulgolem: Maybe we should add `null: false` as well - just so you don't have to fill your head with null options.

karlerikounapuu: Great suggestion! Doing changes.

PR #367, opened 2019-10-16
Wishlist: Add domain_name translation

### LINKED ISSUE #359: Translation error: domeeninimi is already in your wishlist
There is a translation issue with english wishlist new entry error message:

"Domeeninimi is already in your wishlist"

Notice estonian Domeeninimi at the beginning. That part of the message does not come from en translation file https://github.com/internetee/auction_center/blob/master/config/locales/wishlist_items.en.yml

image

karlerikounapuu:
Closes #359

### DIFF
--- a/config/locales/wishlist_items.en.yml
+++ b/config/locales/wishlist_items.en.yml
@@ -19,6 +19,7 @@ en:
   activerecord:
     attributes:
       wishlist_item:
+        domain_name: 'Domain name'
         wishlist: "Wishlist"
     errors:
       models:
--- a/test/system/wishlist_items_test.rb
+++ b/test/system/wishlist_items_test.rb
@@ -66,4 +66,17 @@ class WishlistItemsTest < ApplicationSystemTestCase
     assert_not(page.has_css?('div.notice', text: 'Created successfully.'))
     assert(page.has_text?('Wishlist has too many items'))
   end
+
+  def test_translation_for_already_exist_message
+    @user.update!(locale: 'en')
+
+    fill_in('wishlist_item[domain_name]', with: 'example.test')
+    click_link_or_button('Submit')
+    assert(page.has_text?('Domain name is already in your wishlist'))
+
+    @user.update!(locale: 'et')
+    fill_in('wishlist_item[domain_name]', with: 'example.test')
+    click_link_or_button('Submit')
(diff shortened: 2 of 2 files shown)


config/locales/wishlist_items.en.yml
```diff
@@ -19,6 +19,7 @@ en:
   activerecord:
     attributes:
       wishlist_item:
+        domain_name: 'Domain name'
```
artur-beljajev: I don't think we need to provide English values, since they are kinda translated by Rails itself:

irb(main):003:0> WishlistItem.human_attribute_name :domain_name
=> "Domeeninimi"
irb(main):004:0> I18n.locale
=> :et
irb(main):005:0> I18n.locale = :en
=> :en
irb(main):006:0> WishlistItem.human_attribute_name :domain_name
=> "Domain name"
irb(main):007:0>

test/system/wishlist_items_test.rb
```diff
@@ -66,4 +66,17 @@ def test_must_not_exceed_total_list_capacity
     assert_not(page.has_css?('div.notice', text: 'Created successfully.'))
     assert(page.has_text?('Wishlist has too many items'))
   end
+
+  def test_translation_for_already_exist_message
```
artur-beljajev: It's quite expensive to have system test for every translation string.

karlerikounapuu:
> At the same time I am unable to reproduce the original issue. There must be something wrong with locale setting. I tried to set both `users.locale` and `Rails.configuration.customization['locale']` to `et` (like it is on staging), and still I see the correct translations using both English and Estonian language in UI.
> 
> @karlerikounapuu Perhaps you could provide steps to reproduce?

@artur-beljajev To be honest I wasn't able to produce specific issue myself either. I've got a belief that issue has somehow resolved by itself. Nothing seems to be wrong. Tested with both locales.

Tests were written to reflect that everything is working properly and it's not just me. Safe to close PR and issue I guess. Pushing translation string tests to base doesn't seem needy, I agree.

artur-beljajev:
@vohmar Can you provide the exact steps to reproduce? We're unable to.

vohmar:
cannot reproduce on test nor in production. Seems have resolved somehow. Closing the ticket and PR

PR #369, opened 2019-10-16
Devise: Send an email to user when user password changed

### LINKED ISSUE #167: No password changed email
When user changes their password in the profile they should receive an email about successful password change but no email is sent

karlerikounapuu:
Closes #167

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+28.10.2019
+* Changed due date constraint on invoices table [#386](https://github.com/internetee/auction_center/pull/386)
+
 22.10.2019
 * Fixed multiple error issue in wishlist [#358](https://github.com/internetee/auction_center/issues/358)
 * Long domain names are not cut short any more in full and tablet views [#368](https://github.com/internetee/auction_center/issues/368)
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -120,7 +120,7 @@ Devise.setup do |config|
   # config.send_email_changed_notification = false
 
   # Send a notification email when the user's password is changed.
-  # config.send_password_change_notification = false
+  config.send_password_change_notification = true
 
   # ==> Configuration for :confirmable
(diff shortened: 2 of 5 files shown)


test/system/users/edit_user_test.rb
```diff
@@ -163,4 +163,17 @@ def test_profile_page_has_a_link_to_terms_and_conditions
...
+    visit edit_user_path(@user.uuid)
+    fill_in('user[password]', with: 'newpassword')
+    fill_in('user[password_confirmation]', with: 'newpassword')
+    fill_in('user[current_password]', with: 'password123')
+    assert(page.has_button?('Update'))
+    click_link_or_button('Update')
+    
+    last_email = ActionMailer::Base.deliveries.last
+    assert_equal('Password Changed', last_email.subject)
+    assert_equal(['user@auction.test'], last_email.to)
```
artur-beljajev: In general, such things are tested in mailer tests, but in this particular case I wonder if it's needed to be tested at all, since it is probably already tested by Devise team.

I think the only thing that needs to be tested is that an email is actually sent (that Devise is integrated properly). So I would create an integration test instead with one simple assertion `assert_emails 1`.

karlerikounapuu: Fair enough, good points provided. Creating the provided assertion in edit_user_test seemed logical because edit_user scope bundled with previous assertions seemed right. Recreated as integration test in last commit.

artur-beljajev:
@karlerikounapuu What do you think about https://github.com/internetee/auction_center/pull/369/commits/447159c682357b5c0bded9c77613f2f03f2c0b71?

karlerikounapuu:
@artur-beljajev Naming and test makes more sense in this scenario, yup. I'd give it a go. Mine was just to reflect that devise is indeed sending mails, wasn't even digging into testing an actual flow.

vohmar:
Confirmation email received.
Please rename and squash commits.

PR #370, opened 2019-10-21
Admin: order users from newest to oldest

### LINKED ISSUE #316: Admin: order users from newest to oldest
There is no specified sort order for users list at the moment - the list is displayed in the order the result comes from db that is generally ordered ascending by id meaning that the first users created in the system are always the first in the list.

karlerikounapuu:
Closes #316

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -15,9 +15,9 @@ module Admin
 
     # GET /admin/users
     def index
+      default_params = { 'users.created_at' => 'desc' }
       @users = User.all
-                   .order(orderable_array)
-                   .reorder(created_at: :desc)
+                   .order(orderable_array(default_params))
                    .page(params[:page])
     end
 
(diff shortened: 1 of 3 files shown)


app/controllers/admin/users_controller.rb
```diff
@@ -15,8 +15,9 @@ def new
 
     # GET /admin/users
     def index
+      default_params = { 'users.created_at' => 'desc' }
```
artur-beljajev: Perhaps move to a private method `default_order_params`?

test/system/admin/users/admin_users_list_test.rb
```diff
@@ -74,4 +74,13 @@ def test_certain_fields_are_required
 
     assert_equal(errors_array.to_set, expected_errors.to_set)
   end
+
+  def test_newer_users_are_showed_on_top
+    visit admin_users_path
+    assert_appears_before('user@auction.test', 'administrator@auction.test')
+  end
+
+  def assert_appears_before(earlier_element, later_element)
```
artur-beljajev: Make private?

app/controllers/admin/users_controller.rb
karlerikounapuu: Great suggestion. Done.

PR #374, opened 2019-10-21
Fix wishlist

### LINKED ISSUE #358: Wishlist: multiple errors on failed item entry attempt
If user tries to add an item to the wishlist that is already in their wishlist error "Domeeninimi is already in your wishlist" error is returned. But sometimes there are more than one such messages returned

image

if user goes straight to wishlist view after login only one error message is returned, but if he/she has clicked around between different views the number of error messages is raised.

artur-beljajev:
Fixes #358

### DIFF
--- a/app/javascript/packs/wishlist_items.js
+++ b/app/javascript/packs/wishlist_items.js
@@ -16,6 +16,14 @@ function onlyUnique(value, index, self) {
     return self.indexOf(value) === index;
 }
 
+function clearErrors() {
+    const container = document.getElementById('errors-list');
+
+    while (container.firstChild) {
+        container.removeChild(container.firstChild);
+    }
+}
+
 document.body.addEventListener('ajax:error', (event) => {
     const xhr = event.detail[0];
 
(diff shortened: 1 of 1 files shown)


app/javascript/packs/wishlist_items.js
```diff
@@ -16,7 +16,15 @@ function onlyUnique(value, index, self) {
...
 }
 
-document.addEventListener('ajax:error', (event) => {
+function clearErrors() {
+    const container = document.getElementById('errors-list');
+
+    while (container.firstChild) {
+        container.removeChild(container.firstChild);
+    }
+}
```
karlerikounapuu: As event listener is properly mounted now, why is this part needed though? Only reason I can think of is for a better clarity, since duplicate errors have already vanished.

app/javascript/packs/wishlist_items.js
```diff
@@ -16,7 +16,15 @@ function onlyUnique(value, index, self) {
...
+
+document.body.addEventListener('ajax:error', (event) => {
```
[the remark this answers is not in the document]
artur-beljajev: Great. Let @vohmar test it, and see if everything is fine. It's also interesting, how it works if some another JS request is made, since `ajax:xx` callbacks are triggered for any request, not only for POSTing new wishlist item.

app/javascript/packs/wishlist_items.js
artur-beljajev: I also wanted to get rid of it at first, until I realized that without it every form submit will lead to new errors inserted. So unless they are cleared, they will get appended on and on.

PR #376, opened 2019-10-22
Auctions table: dynamic-width cells, line-break for long domain names

### LINKED ISSUE #368: Long domain names are cut short in tablet view and extra long names in desktop view as well
domain name is most important data in the auction list - not displaying full name is bad for usability.

1. Desktop view -to improve things an idea would be to move reset icon to very end of the table header and  giving it minimum width column and widen the domain name column respectively. This should ´improve things for most long domain names
2. Tablet view - wrap the domain name into multiple lines, stack like in mobile view, or loose some columns like offer actions column (user can access this by clicking on domain name as well)

karlerikounapuu:
Closes #368 
image
image
image

### DIFF
--- a/app/javascript/src/global/_common.scss
+++ b/app/javascript/src/global/_common.scss
@@ -170,10 +170,14 @@ strong {
             @media only screen and (max-width: 767px) {
                 display: none !important;
             }
-        }   
+        }
     }
 }
 
 .ui.table.stackable thead {
     background: #212224;
-  }
+}
+
+.auction-domain-name {
+    word-break: break-all;
+}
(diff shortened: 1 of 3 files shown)


app/javascript/src/global/_common.scss
```diff
@@ -170,10 +170,14 @@ strong {
...
     }
 }
 
 .ui.table.stackable thead {
     background: #212224;
-  }
\ No newline at end of file
+  }
+
+  .auction-domain-name {
```
artur-beljajev: Could you check the indent here. Not sure it is needed.

karlerikounapuu: Checked & fixed

artur-beljajev:
Please squash all commits into single one before merging.

PR #378, opened 2019-10-22
Update .gitignore, add Rubocop/Standard gems to Gemfile

yulgolem:
Does not affect production code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,8 @@
 .rubocop-ruby.yml
 .rubocop.yml
 
+.idea/
+
 # Test files
 /coverage/
 
--- a/Gemfile
+++ b/Gemfile
@@ -33,6 +33,8 @@ end
 group :development do
   gem 'listen', '>= 3.0.5', '< 3.2'
   gem 'pry'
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -33,6 +33,8 @@
 .rubocop-ruby.yml
 .rubocop.yml
 
+.idea/
```
artur-beljajev: Developer-specific files/folders should not be here. How about using local gitignore?

Gemfile
```diff
@@ -33,6 +33,8 @@ end
 group :development do
   gem 'listen', '>= 3.0.5', '< 3.2'
   gem 'pry'
+  gem 'rubocop', '~> 0.75.1', require: false
```
artur-beljajev: Why those gems are needed here?

yulgolem: That's for local linting in RubyMine in dev-time.

artur-beljajev: Doesn't `gem install rubocop` and `rubocop %file%` do the stick?

yulgolem: Hmmm... well, ok, I`ll just decline that.

PR #380, opened 2019-10-22
Add localized terms & conditions setting links

### LINKED ISSUE #372: "Terms and conditions" link should point to localized version of the document
Currently if I use English, the link still points to Estonian version.

https://prnt.sc/pmc74q

It seems we have one https://www.internet.ee/domains/auction-environment-user-agreement.

yulgolem:
Fixes #372
Made this setting project- and locale-agnostic, it may store single URL or hash.


db/structure.sql
```diff
@@ -416,7 +416,7 @@ SET default_tablespace = '';
 SET default_with_oids = false;
 
 --
--- Name: auctions; Type: TABLE; Schema: audit; Owner: -; Tablespace:
```
yulgolem: Should I remove this from PR and leave only migration stamp?

artur-beljajev:
@yulgolem I guess it fixes https://github.com/internetee/auction_center/issues/372?

yulgolem:
> @yulgolem I guess it fixes #372?

Yep. Should I mention it in description?

db/structure.sql
artur-beljajev: Definitely, if this change is not connected to this PR.

artur-beljajev:
> Possible pitfall here is that we don't have :ru locale.

By the way, given the app is not translated to Russian. what's the point of having a TOS link? In what case it will be shows then?

yulgolem:
> > Possible pitfall here is that we don't have :ru locale.
> 
> By the way, given the app is not translated to Russian. what's the point of having a TOS link pointing to Russian version? In what case it will be shown?

No point at all (just saw that the main app has Russian TOS), I've removed it.

karlerikounapuu:
> Can we introduce some general `terms_and_conditions_url` (a container) setting

@artur-beljajev Slightly off-topic, but introducing arrays to Settings as per current design seems to be "hacky" (which is reflecting from my another PR regarding to wishlist). As this functionality is now needed by at least 2 issues, wouldn't it be smart to consider changing our design for current Setting model?

artur-beljajev:
It's kinda hacky, yep. This issue will probably be addresses in a separate issue soon. Thanks for bringing this up!

app/models/setting.rb
```diff
@@ -83,4 +88,16 @@ def self.invoice_reminder_in_days
   def self.wishlist_size
     Setting.find_by(code: :wishlist_size).value.to_i
   end
+
+  class << self
```
karlerikounapuu: Why is new class needed? I'm pretty sure if you write helper functions under `private` in the same Setting class they work as well

yulgolem:
> Perhaps we should test user flow too? Eg correct link is actually present on page when visited? @artur-beljajev ?

Don't think it's needed but added anyway.

app/models/setting.rb
```diff
@@ -83,4 +88,16 @@ def self.invoice_reminder_in_days
...
+    private
+
+    def json_hash?(text)
+      JSON.parse(text).is_a?(Hash)
+    rescue JSON::ParserError
```
artur-beljajev: It's quite dangerous to silently discard an error. Someone (maybe you) will have hard time figuring our why some code doesn't work due to this.

test/models/setting_test.rb
```diff
@@ -63,4 +63,13 @@ def test_class_methods_are_shorthand_for_values
 
     assert_equal(10, Setting.wishlist_size)
   end
+
+  def test_multilocal_terms_and_conditions
+    Setting.find_by(code: :terms_and_conditions_link).update(value: "{\"en\":\"https://example.com\", \"et\":\"https://example.et\"}")
```
artur-beljajev: Fail fast (`update!`). Otherwise it fails silently if `update` does not succeed for some reason.

artur-beljajev: Also, if you really need to supply JSON, why not doing it in a more straightforward fashion: `{ en: 'link' }`?

app/models/setting.rb
yulgolem: I'll modify this part as for #387 .

test/models/setting_test.rb
yulgolem: Well, now here we are testing the saving/validation correctness as well as value parsing.
But deleted I18n-based tests, there is no sense in testing localisation anyway.

PR #381, opened 2019-10-23
Raise PG gem version

### LINKED ISSUE #363: Upgrade to Postgresql 12

yulgolem:
Checked on PG v 12.


vohmar:
I removed the #363 reference from description as this PR does not actually close the issue. Here we update gem version, but the db version is still 9.x.

artur-beljajev:
Updated original issue (the correct version to upgrade to is 11, not 12)

artur-beljajev:
It's indeed better to decouple `pg` gem and Postgresql update/upgrade, since the latter requires infrastructure change.

Issue #321, opened 2019-08-13
Unexpected sorting behavior at `/admin/auctions`


artur-beljajev:
Only the first sorting column is taken into account since the page load.

Example:
Navigate to `/admin/auctions?order%5Badmin_auction_decorators.domain_name%5D=asc` and try to sort by any other column than domain name. Notice that auctions are only sorted by domain name, with no regard to other columns, specified in URL (`/admin/auctions?order%5Badmin_auction_decorators.domain_name%5D=asc&order%5Badmin_auction_decorators.ends_at%5D=asc&order%5Badmin_auction_decorators.starts_at%5D=asc`).

I am not sure what is the expected behavior here, but it's definitely not good when I click on multiple columns and don't get expected results nor any feedback about improper use of UI.

yulgolem:
I suppose works as expected.

I've added a fixture with same domain name to check.
Unsorted list looks like:
![image](https://user-images.githubusercontent.com/7688888/67462681-7d868180-f659-11e9-8193-75a9c3ed2a1c.png)

Then sorted by domains name desc:
![image](https://user-images.githubusercontent.com/7688888/67462775-9ee76d80-f659-11e9-9d93-444ba76600f4.png)
Note the dates of start/end and # of offers on two last auctions with same domain.
The order part of SQL-query now is `ORDER BY auctions.domain_name asc NULLS LAST `.

Now let's sort by # of offers asc:
![image](https://user-images.githubusercontent.com/7688888/67464054-10c0b680-f65c-11e9-9e73-dac4a8f6bae6.png)
Again note the dates of start/end and # of offres on two last auctions with same domain.
And the order part of query is ` ORDER BY auctions.domain_name asc NULLS LAST, auctions.number_of_offers desc NULLS LAST`.

So - the next level of sorting is on only when there is a some equal entities of the previous level - ie domain name in our case. If we'll try to sort this dataset bu some other field - the sorting will fails since its already sorted.

That was locally, I'll try and see how it's going on staging.

vohmar:
closing this issue with partial solution by PR #383. Work is continued in issue #400 which has more details on the topic

PR #383, opened 2019-10-24
Set zero-sum in auction index to NaN (i18n-ized)

yulgolem:
Closes sorting inconvenience without data in #321

### DIFF
--- a/app/views/admin/auctions/index.html.erb
+++ b/app/views/admin/auctions/index.html.erb
@@ -61,7 +61,7 @@
                                 </td>
                                 <td><%= I18n.l(auction.starts_at) %></td>
                                 <td><%= I18n.l(auction.ends_at) %></td>
-                                <td class="auction-highest-offer"><%= auction.highest_price %></td>
+                                <td class="auction-highest-offer"><%= auction.highest_offer_id.present? ? auction.highest_price : t('.no_price')%></td>
                                 <td class="auction-offers-count"><%= auction.number_of_offers %></td>
                                 <td></td>
                             </tr>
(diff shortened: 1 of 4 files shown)


app/views/admin/auctions/index.html.erb
```diff
@@ -61,7 +61,7 @@
                                 </td>
                                 <td><%= I18n.l(auction.starts_at) %></td>
                                 <td><%= I18n.l(auction.ends_at) %></td>
-                                <td class="auction-highest-offer"><%= auction.highest_price %></td>
+                                <td class="auction-highest-offer"><%= auction.highest_offer_id.present? ? auction.highest_price : t('nan')%></td>
```
artur-beljajev: - How about `t('.')` shortcut to avoid cluttering global namespace?
- Is there some more meaningful and abstract name for this? Something that would reveal the intention, like `no_price`?

test/system/admin/admin_auctions_test.rb
```diff
@@ -113,7 +113,7 @@ def test_administrator_can_see_all_created_auctions
     assert(page.has_link?('expired.test', href: admin_auction_path(@expired_auction.id)))
 
     prices = page.find_all('.auction-highest-offer').map(&:text)
-    assert_equal(['50.00', '10.00', '0.00', '100.00'].to_set, prices.to_set)
+    assert_equal(['50.00', '10.00', I18n.t('nan'), '100.00'].to_set, prices.to_set)
```
artur-beljajev: Perhaps it's better to hardcode this? If it's not translated, `nil` will be returned, and I am not sure how this will behave then.
I wonder whether it's really a system test. Can this logic be tested in unit tests?

yulgolem: I suppose it's a system test indeed - we might want to see if view behaves as expected and shows 'NaN' if there is no data.
But hardcode is a thing thought, made it so.

artur-beljajev:
Do you have a plan to squash commits?

yulgolem:
> @yulgolem Do you have a plan to squash commits?

Sure thing.

Issue #271, opened 2019-05-21
Wishlist: automatic top level domain extension


vohmar:
Nice to have feature:
* if user enters domainname without tld extension it should be added automatically (siliarly to rest-whois)
* the list of allowed tld extensions should be limited and configurable - in our case the list is .ee (default for previous point), .pri.ee, .com.ee, .med.ee, .fie.ee. If user enters anything else error should be returne (en: Domain name is invalid; et: Domeeninimi ei ole korrektne)

PR #384, opened 2019-10-24
Wishlist: Allow only domains that have whitelisted domain extension

karlerikounapuu:
Resolves second pivot of #271

### DIFF
--- a/app/helpers/settings_helper.rb
+++ b/app/helpers/settings_helper.rb
@@ -1,0 +1,18 @@
+module SettingsHelper
+  def formatted_setting_value(setting)
+    return setting unless filled_json_array?(setting)
+
+    elements = ''
+    values = JSON.parse(setting)
+    values.each do |value|
+      elements << content_tag(:a, value, class: 'ui label')
+    end
+    simple_format(elements)
+  end
+
+  def filled_json_array?(text)
+    JSON.parse(text).is_a?(Array) && JSON.parse(text).any?
+  rescue JSON::ParserError
+    false
+  end
+end
--- a/app/models/setting.rb
+++ b/app/models/setting.rb
@@ -1,9 +1,8 @@
 class Setting < ApplicationRecord
   include Concerns::FormatValidator
-  validates :code, presence: true
+  validates :code, presence: true, uniqueness: true
   validates :description, presence: true
   validates :value, presence: true
-  validates :code, uniqueness: true
 
   def self.auction_currency
     Setting.find_by(code: :auction_currency).value
@@ -101,5 +100,10 @@ class Setting < ApplicationRecord
   def self.violations_count_regulations_link
     hash = Setting.find_by(code: :violations_count_regulations_link)&.value
     hash.present? ? JSON.parse(hash).with_indifferent_access[I18n.locale] : nil
+  end
+
+  def self.wishlist_supported_domain_extensions
+    extensions = Setting.find_by(code: :wishlist_supported_domain_extensions)
+    extensions.present? ? JSON.parse(extensions.value) : []
   end
 end
--- a/app/models/wishlist_item.rb
+++ b/app/models/wishlist_item.rb
@@ -6,16 +6,22 @@ class WishlistItem < ApplicationRecord
   before_validation :set_domain_name_to_unicode
 
   validates :domain_name, uniqueness: { scope: :user_id }
+
+  # 1) Allows 1 to 61 characters with Estonian diacritic signs as domain name / third-level domain.
+  # 2) 1 to 61 chars with Estonian diacritic signs as second-level domain (optional) and 1 to 61
+  # characters as top level domain, which represents domain name extension when concatenated.
+  domain_name_regexp = /\A[a-z0-9\-\u00E4\u00F5\u00F6\u00FC\u0161\u017E]{1,61}\.
+                       ([a-z0-9\-\u00E4\u00F5\u00F6\u00FC\u0161\u017E]{1,61}\.)?[a-z0-9]{1,61}\z/x
+
   validates :domain_name,
             presence: true,
             allow_blank: false,
             format: {
-              # allows 1 to 61 characters with Estonian diacritic signs and 1 to 61 character
-              # of the top level domain.
-              with: /\A[a-z0-9\-\u00E4\u00F5\u00F6\u00FC\u0161\u017E]{1,61}\.[a-z0-9]{1,61}\z/,
+              with: domain_name_regexp,
             }
 
   validate :must_fit_in_wishlist_size, on: :create
+  validate :valid_domain_extension, on: :create
 
   scope :for_user, ->(user_id) { where(user_id: user_id) }
 
@@ -43,5 +49,15 @@ class WishlistItem < ApplicationRecord
     else
       0
     end
+  end
+
+  private
+
+  # Validate that FQDN has supported extension
+  def valid_domain_extension
+    return if Setting.wishlist_supported_domain_extensions.empty?
+    return if Setting.wishlist_supported_domain_extensions.include?(domain_name.split('.', 2).last)
+
+    errors.add(:domain_name, :invalid) if errors[:domain_name].blank?
   end
 end
--- a/app/views/admin/settings/index.html.erb
+++ b/app/views/admin/settings/index.html.erb
@@ -22,7 +22,7 @@
                                 <%=link_to setting.code, admin_setting_path(setting) %>
                             </td>
                             <td><%= setting.description %></td>
-                            <td><%= setting.value %></td>
+                            <td><%= formatted_setting_value(setting.value) %></td>
                         </tr>
                     <% end %>
                 </tbody>
--- a/app/views/admin/settings/show.html.erb
+++ b/app/views/admin/settings/show.html.erb
@@ -15,7 +15,7 @@
 
             <div class="item">
                 <div class="header"><%= t('settings.value') %></div>
-                <%= @setting.value %>
+                <%= formatted_setting_value(@setting.value) %>
             </div>
 
             <div class="item">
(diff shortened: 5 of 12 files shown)


lib/tasks/wishlist.rake
```diff
@@ -0,0 +1,21 @@
...
+      Supported domain extensions for wishlist domain monitoring.
+    TEXT
+
+    extensions = ['ee', 'pri.ee', 'com.ee', 'med.ee', 'fie.ee']
+    domain_extensions = Setting.new(
+      code: :wishlist_supported_domain_extensions,
+      value: extensions,
+      description: domain_extensions_description
+    )
+    if domain_extensions.save
```
artur-beljajev: Fail fast (`save!`) approach might be better here. This way you get meaningful description right away instead of just "Failed", if something goes wrong. `puts` might be eliminated entirely in this case.

Also I think it's important to denote that is is a temporary data migration, by putting it under "data_migrations" namespace (and folder), for example.

app/models/setting.rb
```diff
@@ -83,4 +84,38 @@ def self.invoice_reminder_in_days
...
+  end
+end
+
+private
+
+def valid_format_for_wishlist_domain_extensions
+  return unless code == 'wishlist_supported_domain_extensions'
+  return if valid_domain_extension_elements?(value)
+
+  errors.add(:value, 'is invalid')
```
artur-beljajev: `:invalid` allows it to be localized, instead of being hardcoded.

app/models/setting.rb
```diff
@@ -83,4 +84,38 @@ def self.invoice_reminder_in_days
...
+rescue JSON::ParserError
+  false
+rescue NoMethodError
+  false
+end
+
+def valid_domain_extension_elements?(hash)
+  return false unless json_array?(hash)
+
+  r = /\A([a-z0-9\-\u00E4\u00F5\u00F6\u00FC\u0161\u017E]{1,61}\.)?[a-z0-9]{1,61}\z/x
```
artur-beljajev: Not sure I understand this.

app/models/wishlist_item.rb
```diff
@@ -12,10 +12,12 @@ class WishlistItem < ApplicationRecord
             format: {
               # allows 1 to 61 characters with Estonian diacritic signs and 1 to 61 character
               # of the top level domain.
-              with: /\A[a-z0-9\-\u00E4\u00F5\u00F6\u00FC\u0161\u017E]{1,61}\.[a-z0-9]{1,61}\z/,
+              with: /\A[a-z0-9\-\u00E4\u00F5\u00F6\u00FC\u0161\u017E]{1,61}\.
```
artur-beljajev: Is it the same regexp as on https://github.com/internetee/auction_center/pull/384/files#diff-8c3f2a547e7f7d66e1756971a1ef20c2R117?

app/models/wishlist_item.rb
```diff
@@ -44,4 +46,12 @@ def number_of_items_for_user
       0
     end
   end
+
+  # Validate that FQDN has supported extension
+  def valid_domain_extension
+    return if Setting.wishlist_supported_domain_extensions.empty?
+    return if Setting.wishlist_supported_domain_extensions.include?(domain_name.split('.', 2).last)
+
+    errors.add(:domain_name, I18n.t('is_invalid')) if errors[:domain_name].blank?
```
artur-beljajev: Same as above (`:invalid`).

app/views/admin/settings/index.html.erb
```diff
@@ -22,7 +22,16 @@
                                 <%=link_to setting.code, admin_setting_path(setting) %>
                             </td>
                             <td><%= setting.description %></td>
-                            <td><%= setting.value %></td>
+                            <td>
+                              <% if filled_json_array?(setting.value)
+                                values = JSON.parse(setting.value)
+                                values.each do |value| %>
+                                  <a class='ui label'><%= value %></a>
+                              <% end
```
artur-beljajev: Does this work this way? (`else` without `if`?)
Can we extract logic from view?

test/models/wishlist_item_test.rb
```diff
@@ -72,4 +72,31 @@ def test_domain_name_must_be_a_valid_domain
     assert_not(item.valid?)
     assert_equal(['is invalid'], item.errors[:domain_name])
   end
+
+  def test_domain_has_valid_extension
+    supported_extensions = ["ee", "pri.ee", "com.ee", "med.ee", "fie.ee"]
```
artur-beljajev: Do they all really needed here? What if we create a single one and leave only two tests - `valid` and `invalid`?

app/views/admin/settings/index.html.erb
karlerikounapuu: Actually `if` was present, but doesn't really matter because I'm moving whole view logic to helper instead.

app/models/setting.rb
karlerikounapuu: Regex to check that domain extension is in valid format, without any whitespaces. Accepts xxx / xxx.xx

app/views/admin/settings/index.html.erb
artur-beljajev: But it was closed right before `else` https://github.com/internetee/auction_center/blob/6f6215a30828d46b816dbfc90635fa44609fee34/app/views/admin/settings/index.html.erb#L30

app/models/wishlist_item.rb
karlerikounapuu: No, it's not. The one in `Setting` is limited to one subdomain (ee, pri.ee). `WishlistItem` regex doesn't have such limitation.

app/views/admin/settings/index.html.erb
karlerikounapuu: Internal `do` loop was closed with that `end`, but indentation was wrong, indeed. Seems like `html.erb` doesn't really care about that.

app/models/setting.rb
```diff
@@ -3,6 +3,7 @@ class Setting < ApplicationRecord
   validates :description, presence: true
   validates :value, presence: true
   validates :code, uniqueness: true
+  validate :valid_format_for_wishlist_domain_extensions, on: %i[create update]
```
artur-beljajev: Are there any other contexts besides `create` and `update`?

app/helpers/settings_helper.rb
```diff
@@ -0,0 +1,20 @@
...
+      elements << content_tag(:a, value, class: 'ui label')
+    end
+    simple_format(elements)
+  end
+
+  def filled_json_array?(text)
+    JSON.parse(text).is_a?(Array) && JSON.parse(text).first.present?
+  rescue JSON::ParserError
+    false
+  rescue NoMethodError
```
artur-beljajev: Why these `rescue` are needed? There must be a strong reason to catch an exception, since you basically suppress an error here.

app/models/setting.rb
```diff
@@ -83,4 +84,38 @@ def self.invoice_reminder_in_days
...
   end
+
+  def self.wishlist_supported_domain_extensions
+    extensions = Setting.find_by(code: :wishlist_supported_domain_extensions)
+    if extensions.present?
+      return JSON.parse(extensions.value) if json_array?(extensions.value)
+    end
+    []
+  end
+end
```
artur-beljajev: `Setting` class is closed here?

app/models/setting.rb
```diff
@@ -83,4 +84,38 @@ def self.invoice_reminder_in_days
   def self.wishlist_size
     Setting.find_by(code: :wishlist_size).value.to_i
   end
+
+  def self.wishlist_supported_domain_extensions
```
artur-beljajev: This method seems quite uncertain (too many `if`s). In what case the setting is absent? It seems `return...` is the mainline logic here, so how about moving it to the end of the method and `return []` before it? (Guard Clause https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html)?

app/models/setting.rb
artur-beljajev: Can we document this knowledge in the variable name? Like `domain_name_regexp`? If not, it's worth commenting it describing what does it match. Maybe like here https://github.com/internetee/auction_center/blob/df6009fbc588cf7605094e35b845100434c3ddab/app/models/wishlist_item.rb#L13 or in some other suitable way.

app/models/wishlist_item.rb
artur-beljajev: Same as above: can we extract this regexp somewhere and name it based on its semantic meaning?

app/views/admin/settings/show.html.erb
```diff
@@ -15,7 +15,14 @@
 
             <div class="item">
                 <div class="header"><%= t('settings.value') %></div>
-                <%= @setting.value %>
+                  <% if filled_json_array?(@setting.value)
```
artur-beljajev: Can this logic be extracted from view?

test/models/setting_test.rb
```diff
@@ -63,4 +63,20 @@ def test_class_methods_are_shorthand_for_values
 
     assert_equal(10, Setting.wishlist_size)
   end
+
+  def test_valid_format_for_wishlist_domain_extension_setting
+    @subject = settings(:application_name)
```
artur-beljajev: What does this `subject` stand for? Perhaps `setting` is more readable?

app/models/setting.rb
karlerikounapuu: `create` and `update` are only contexts that require this validation since other methods don't manipulate with data

app/models/setting.rb
karlerikounapuu: Getting fixed.

app/views/admin/settings/show.html.erb
karlerikounapuu: Missed this one. Thanks!

app/helpers/settings_helper.rb
```diff
@@ -0,0 +1,18 @@
+module SettingsHelper
```
yulgolem: Consider using [this concern](https://github.com/internetee/auction_center/blob/master/app/models/concerns/format_validator.rb) , it's in `master` now.

PR #387, opened 2019-10-28
229 hide domain names in ban messages

### LINKED ISSUE #229: Hide references to domain names from ban messages if these are not being auctioned
Current soft ban message lists the domain names that user is banned from participating. The ban should last until the specific domain is out of auction cycle ie no bids were made and domain is released or an auction winner has registered it.

If this happens, the ban is not relevant any more and hence the ui should not return that domain name. So the auction center should remove the domain name from the presented list if the domain is not at auction. If the removed domain name is the last the whole ban message should be removed.

Referencing: https://github.com/internetee/auction_center/issues/202

yulgolem:
Closes #229 .

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -21,6 +21,16 @@ module ApplicationHelper
     end
   end
 
+  def strikes_banner
+    return unless session['auction.bans']
+    domains = session['auction.bans'][0]
+    message = strikes_message(domains)
+
+    content_tag(:div, class: 'ui message strikes') do
+      content_tag(:div, message, class: 'header')
+    end
+  end
+
   def banned_banner
     return unless session['auction.bans']
 
@@ -39,6 +49,14 @@ module ApplicationHelper
       t(:banned_completely, valid_until: valid_until.to_date)
     else
       t(:banned, domain_names: domains.join(', '))
+    end
+  end
+
+  def strikes_message(domains)
+    if domains.count
+      "You have got #{domains.count} of 3 bans last year."
+    else
+      "You have got 3 of 3 bans last year."
     end
   end
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -48,16 +48,22 @@ module ApplicationHelper
     if valid_until
       t(:banned_completely, valid_until: valid_until.to_date)
     else
-      t(:banned, domain_names: domains.join(', '))
+      active_domains = check_active_auctions_for(domains)
+      t(:banned, domain_names: active_domains.join(', '))
     end
   end
 
   def strikes_message(domains)
-    if domains.count
+    if domains
       "You have got #{domains.count} of 3 bans last year."
     else
       "You have got 3 of 3 bans last year."
     end
+  end
+
+  def check_active_auctions_for(domains)
+    active_auction_domains = Auction.active.map(&:domain_name)
+    domains.reject{ |element| !active_auction_domains.include?(element) }
   end
 
   def links(links_list)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -40,10 +40,14 @@ module ApplicationHelper
 
   def ban_error_message(domains, valid_until)
     if valid_until
-      t(:banned_completely, valid_until: valid_until.to_date)
+      t('auctions.banned_completely', valid_until: valid_until.to_date)
     else
       active_domains = check_active_auctions_for(domains)
-      active_domains.count.positive? ? t(:banned, domain_names: active_domains.join(', ')) : ''
+      if active_domains.count.positive?
+        I18n.t('auctions.banned', domain_names: active_domains.join(', '))
+      else
+        ''
+      end
     end
   end
 
@@ -54,9 +58,9 @@ module ApplicationHelper
 
   def violation_message(domains_count)
     link = Setting.violations_count_regulations_link
-    t('violation_message_html', violations_count_regulations_link: link,
-                                violations_count: domains_count,
-                                ban_number_of_strikes: Setting.ban_number_of_strikes)
+    t('auctions.violation_message_html', violations_count_regulations_link: link,
+                                         violations_count: domains_count,
+                                         ban_number_of_strikes: Setting.ban_number_of_strikes)
   end
 
   def links(links_list)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -21,16 +21,6 @@ module ApplicationHelper
     end
   end
 
-  def strikes_banner
-    return unless session['auction.bans']
-    domains = session['auction.bans'][0]
-    message = strikes_message(domains)
-
-    content_tag(:div, class: 'ui message strikes') do
-      content_tag(:div, message, class: 'header')
-    end
-  end
-
   def banned_banner
     return unless session['auction.bans']
 
@@ -49,15 +39,8 @@ module ApplicationHelper
       t(:banned_completely, valid_until: valid_until.to_date)
     else
       active_domains = check_active_auctions_for(domains)
-      t(:banned, domain_names: active_domains.join(', '))
-    end
-  end
-
-  def strikes_message(domains)
-    if domains
-      "You have got #{domains.count} of 3 bans last year."
-    else
-      "You have got 3 of 3 bans last year."
+      result = active_domains.count.positive? ? t(:banned, domain_names: active_domains.join(', ')) : ''
+      result << t(:total_strikes, count: domains.count)
     end
   end
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -30,7 +30,7 @@ module ApplicationHelper
     content_tag(:div, class: 'ui message ban') do
       result = content_tag(:div, message, class: 'header')
       if domains && domains.count < Setting.ban_number_of_strikes
-        result << content_tag(:p, violation_message(domains.count))
+        result << content_tag(:p, violation_message(domains.count), class: 'violation-message')
       end
       result
     end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -28,7 +28,11 @@ module ApplicationHelper
     message = ban_error_message(domains, valid_until)
 
     content_tag(:div, class: 'ui message ban') do
-      content_tag(:div, message, class: 'header')
+      result = content_tag(:div, message, class: 'header')
+      if domains && domains.count < Setting.ban_number_of_strikes
+        result << content_tag(:p, violation_message(domains.count))
+      end
+      result
     end
   end
 
@@ -39,14 +43,20 @@ module ApplicationHelper
       t(:banned_completely, valid_until: valid_until.to_date)
     else
       active_domains = check_active_auctions_for(domains)
-      result = active_domains.count.positive? ? t(:banned, domain_names: active_domains.join(', ')) : ''
-      result << t(:total_strikes, count: domains.count)
+      active_domains.count.positive? ? t(:banned, domain_names: active_domains.join(', ')) : ''
     end
   end
 
   def check_active_auctions_for(domains)
-    active_auction_domains = Auction.active.map(&:domain_name)
-    domains.reject{ |element| !active_auction_domains.include?(element) }
+    active_auction_domains = Auction.active.pluck(:domain_name)
+    domains.reject { |element| !active_auction_domains.include?(element) }
+  end
+
+  def violation_message(domains_count)
+    link = Setting.violations_count_regulations_link
+    t('violation_message_html', violations_count_regulations_link: link,
+                                violations_count: domains_count,
+                                ban_number_of_strikes: Setting.ban_number_of_strikes)
   end
 
   def links(links_list)
--- a/app/models/setting.rb
+++ b/app/models/setting.rb
@@ -4,11 +4,6 @@ class Setting < ApplicationRecord
   validates :description, presence: true
   validates :value, presence: true
   validates :code, uniqueness: true
-  validate :validate_value
-
-  FORMAT_VALIDATIONS = {
-    violations_count_regulations_link: [:validate_json_hash],
-  }.with_indifferent_access.freeze
 
   def self.auction_currency
     Setting.find_by(code: :auction_currency).value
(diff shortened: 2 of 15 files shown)


app/helpers/application_helper.rb
```diff
@@ -38,10 +42,23 @@ def ban_error_message(domains, valid_until)
     if valid_until
       t(:banned_completely, valid_until: valid_until.to_date)
     else
-      t(:banned, domain_names: domains.join(', '))
+      active_domains = check_active_auctions_for(domains)
+      active_domains.count.positive? ? t(:banned, domain_names: active_domains.join(', ')) : ''
```
artur-beljajev: Avoid cluttering global namespace (`en.banned`).

app/helpers/application_helper.rb
```diff
@@ -28,7 +28,11 @@ def banned_banner
     message = ban_error_message(domains, valid_until)
 
     content_tag(:div, class: 'ui message ban') do
-      content_tag(:div, message, class: 'header')
+      result = content_tag(:div, message, class: 'header')
+      if domains && domains.count < Setting.ban_number_of_strikes
+        result << content_tag(:p, violation_message(domains.count))
+      end
+      result
```
artur-beljajev: Any chance to avoid using helpers and using pure objects instead?

app/models/setting.rb
```diff
@@ -83,4 +83,23 @@ def self.invoice_reminder_in_days
...
+    else
+      value.to_s
+    end
+  end
+
+  def self.json_hash?(text)
+    JSON.parse(text).is_a?(Hash)
+  rescue JSON::ParserError
+    false
+  rescue NoMethodError
```
artur-beljajev: Such things usually lead to debugging nightmare, since you're effectively suppressing the errors. How about revamping the logic so that it doesn't require `rescue`ing? Perhaps it can be substituted by guard clause?

app/models/setting.rb
```diff
@@ -83,4 +83,23 @@ def self.invoice_reminder_in_days
...
+    false
+  end
+
+  private_class_method :json_hash?
```
artur-beljajev: If we really need class method (can we avoid it?), `class << self` might be a more versatile way to have private class method.

db/seeds.rb
```diff
@@ -131,6 +131,23 @@
...
+  or absolute ('https://example.com/terms_and_conditions.pdf'). Relative URL must start with a
+  forward slash.
+  Default: https://example.com#some_anchor
+TEXT
+
+violations_count_regulations_setting = Setting.new(code: :violations_count_regulations_link,
+                                            value: "https://example.com#some_anchor",
+                                            description: violations_count_regulations_description)
+
+violations_count_regulations_setting.save
```
artur-beljajev: `save!` might save debugging time if something goes wrong and a record will not be saved.

lib/tasks/data_migrations/violations_count_regulations_link_setting.rake
```diff
@@ -0,0 +1,26 @@
...
+        or absolute ('https://example.com/terms_and_conditions.pdf'). Relative URL must start with a
+        forward slash.
+        Default: https://example.com#some_anchor
+      TEXT
+
+      violations_count_regulations_setting = Setting.new(code: :violations_count_regulations_link,
+                                                         value: "https://example.com#some_anchor",
+                                                         description: violations_count_regulations_description)
+
+      violations_count_regulations_setting.save
```
artur-beljajev: Fail fast is a must here (bang version - `save!`). Otherwise it just silently fails and you need to spend time debugging it.

lib/tasks/data_migrations/violations_count_regulations_link_setting.rake
```diff
@@ -0,0 +1,26 @@
...
+
+    end
+
+    puts "Violation count regulation link setting updated"
```
artur-beljajev: This makes sense only if above mentioned thing is in place. Otherwise it says "OK, nothing goes wrong", whereas in fact nothing was saved (due to validation, callbacks, whatever).

test/models/setting_test.rb
```diff
@@ -63,4 +63,13 @@ def test_class_methods_are_shorthand_for_values
 
     assert_equal(10, Setting.wishlist_size)
   end
+
+  def test_multilocale_violations_count_regulations_link
+    Setting.find_by(code: :violations_count_regulations_link).update(value: "{\"en\":\"https://example.com#some_anchor\", \"et\":\"https://example.et#some_anchor\"}")
```
artur-beljajev: `update!`, same reason as above.

test/models/setting_test.rb
```diff
@@ -63,4 +63,13 @@ def test_class_methods_are_shorthand_for_values
...
+    I18n.with_locale(:et) { assert_equal('https://example.et#some_anchor', Setting.violations_count_regulations_link) }
+    I18n.with_locale(:en) { assert_equal('https://example.com#some_anchor', Setting.violations_count_regulations_link) }
+    Setting.find_by(code: :violations_count_regulations_link).update(value: "https://alphabetagamma.com#some_anchor")
+    I18n.with_locale(:et) { assert_equal('https://alphabetagamma.com#some_anchor', Setting.violations_count_regulations_link) }
+    I18n.with_locale(:en) { assert_equal('https://alphabetagamma.com#some_anchor', Setting.violations_count_regulations_link) }
```
artur-beljajev: Looks cumbersome. I guess there is no other way given the current implementation of settings?

test/system/bans_test.rb
```diff
@@ -91,20 +91,35 @@ def test_banned_user_can_see_the_ban_notification_for_one_domain
...
     TEXT
 
     visit auctions_path
     assert(page.has_css?('div.ban', text: text))
   end
 
+  def test_banned_user_can_see_the_ban_notification_only_for_active_auctions
+    Ban.create!(user: @participant,
+                domain_name: 'dummy-domain.net',
+                valid_from: Time.zone.today - 1, valid_until: Time.zone.today + 5)
```
artur-beljajev: Magic numbers. The intention is not clear. Why just 5? You might need to use `travel_to`, otherwise sporadic errors might appear due to variable current time.

test/system/bans_test.rb
```diff
@@ -119,6 +134,7 @@ def test_banned_user_can_see_the_ban_notification_for_longest_repetitive_ban
 
     visit auctions_path
     assert(page.has_css?('div.ban', text: text))
+    assert_not(page.has_css?('div.ban > p'))
```
artur-beljajev: `has_no_css?`

app/helpers/application_helper.rb
yulgolem: Moved all the translations from here to 'auctions.*'

lib/tasks/data_migrations/violations_count_regulations_link_setting.rake
yulgolem: Moved this to the end of transaction.

test/models/setting_test.rb
yulgolem: Well, there is no need to test localisation, so let's test just if setting returns string or value of a hash k/v pair correctly.

app/helpers/application_helper.rb
yulgolem: I prefer not to - we are already in helper using helpers here, so I prefer to keep the style constant.

test/system/bans_test.rb
yulgolem: We are already traveling to in the setup phase, these numbers are just to get a range. Moved them to named constants anyway.

test/system/bans_test.rb
yulgolem: Other tests here uses assert_not method, so I prefer to keep the codestyle constant.

artur-beljajev: `has_no_css?` is much more readable and more simple, than `assert_not(page.has_css?)`.

yulgolem: Well, here we would like to check what given block doesn't shown to user at all, so check for text maybe wouldn't the best idea, but for absence of css is.
Changed to has_no_css? .

app/models/setting.rb
```diff
@@ -83,4 +84,24 @@ def self.invoice_reminder_in_days
...
+  def valid_violations_count_regulations_link_format
+    return unless code == 'violations_count_regulations_link'
+    return if valid_violations_count_json?(value)
+
+    errors.add(:value, :invalid)
+  end
+
+  def valid_violations_count_json?(value)
+    JSON.parse(value).is_a?(Hash)
+  rescue JSON::ParserError
```
artur-beljajev: Avoid failing silently.

yulgolem: Hmmm, maybe I should save  exception message in validation error?
I'll do so anyway.

artur-beljajev: Do not catch without a serious reason, or it will lead to hard-to-find errors and long debugging sessions. Let it fail. In this case I would rethink  the design. The main question I had, was "Why should it fail while parsing?". If it's totally unexpected, let it fail. If it is expected, something is wrong with the design, in my opinion.

test/system/bans_test.rb
```diff
@@ -91,20 +94,35 @@ def test_banned_user_can_see_the_ban_notification_for_one_domain
...
   end
 
+  def test_banned_user_can_see_the_ban_notification_only_for_active_auctions
+    Ban.create!(user: @participant,
+                domain_name: 'dummy-domain.net',
+                valid_from: Time.zone.today - DAYS_BEFORE, valid_until: Time.zone.today + DAYS_AFTER)
+
+    sign_in_manually
+
+    text = <<~TEXT.squish
```
artur-beljajev: `text` is too broad. It is probably a notification? Or rather `expected_notification`? (if it's used only in `assert`).

test/system/bans_test.rb
```diff
@@ -91,20 +94,35 @@ def test_banned_user_can_see_the_ban_notification_for_one_domain
...
+      Number of current violations: 2 of 3
+    TEXT
+
+    visit auctions_path
+    assert(page.has_css?('div.ban > p', text: text))
```
artur-beljajev: Why not just `assert_text expected_notification` (see comment above)?

test/system/bans_test.rb
```diff
@@ -117,8 +135,10 @@ def test_banned_user_can_see_the_ban_notification_for_longest_repetitive_ban
       invoices until 2010-07-10.
     TEXT
 
+
     visit auctions_path
     assert(page.has_css?('div.ban', text: text))
+    has_no_css?('div.ban > p')
```
artur-beljajev: The intention is not clear. Consider testing it from a user viewpoint.

test/models/setting_test.rb
```diff
@@ -63,4 +63,9 @@ def test_class_methods_are_shorthand_for_values
 
     assert_equal(10, Setting.wishlist_size)
   end
+
+  def test_multilocale_violations_count_regulations_link
+    Setting.find_by(code: :violations_count_regulations_link).update!(value: "{\"en\":\"https://example.com#some_anchor\", \"et\":\"https://example.et#some_anchor\"}")
```
artur-beljajev: It seems `et` key has no effect on the test. Can we remove it to avoid one having such assumption?

Also, how about implying value `https://example.com#some_anchor` to say `https://regulations.test`, extracting it to a variable `regulations_link` and reusing it, making a test more readable? (you clearly see the connection between input and output).

app/models/setting.rb
```diff
@@ -83,4 +88,32 @@ def self.invoice_reminder_in_days
...
+
+    methods.each do |method|
+      send(method)
+    end
+  end
+
+  def validate_json_hash
+    return if parsable_json_hash?(value)
+
+    errors.add(:value, :invalid)
```
artur-beljajev: It should be `errors.add(:value, :invalid) unless parsable_json_hash?(value)`. This is the mainline of code, it should not be a guard clause.

app/models/setting.rb
```diff
@@ -83,4 +88,32 @@ def self.invoice_reminder_in_days
...
+  end
+
+  private
+
+  def validate_value
+    methods = FORMAT_VALIDATIONS[code]
+    return unless methods
+
+    methods.each do |method|
+      send(method)
```
artur-beljajev: Why `send` is needed here? This looks magical and doesn't communicate what this code does and why at a high level.

artur-beljajev:
@yulgolem 
> codeclimate/total-coverage — 96% (-0.2% change)

We can't do anything about it?

test/system/bans_test.rb
yulgolem: Precisely from user's viewpoint we are testing here
1) user can see the text message on total ban
2) message on violations count didn't displayed.

So I'd rather keep it that way.

artur-beljajev: A user doesn't see HTML code, still :). This test is fragile, since it relies on HTML structure too heavily.

> message on violations count didn't displayed.

Can we document this in code?

Like `assert_no_text 'message on violations'`?

test/models/setting_test.rb
yulgolem: I'd rather keep that key - it helps us understand that not only hash is saved correctly, but we are getting correct key/value pair from hash.
But changing value to more info is a good idea.

artur-beljajev: > but we are getting correct key/value pair from hash.

Where is it tested, I don't get. Isn't it a concern of unit test, not system?

app/models/setting.rb
yulgolem: Renamed variables here to make more sense.

yulgolem:
> @yulgolem
> 
> > codeclimate/total-coverage — 96% (-0.2% change)
> 
> We can't do anything about it? (note `-0.2%`).

Rebased from master, Karl had there a PR on this.

app/models/setting.rb
```diff
@@ -83,4 +88,30 @@ def self.invoice_reminder_in_days
...
+    hash.present? ? JSON.parse(hash).with_indifferent_access[I18n.locale] : nil
+  end
+
+  private
+
+  def validate_value
+    validation_methods_for_code = FORMAT_VALIDATIONS[code]
+    return unless validation_methods_for_code
+
+    validation_methods_for_code.each do |validation_method|
```
artur-beljajev: What is the disadvantage of replacing meta programming with simple method calls? The disadvantage of meta programming is the complexity it introduces. For such simple goal it seems like overkill.

yulgolem: When we'll add Karl's and another mine PR on Setting with this methods, all we need to do will just to add validator method names to FORMAT_VALIDATIONS hash.
Yes, right here it might be overkill, but there is two more PRs with same functionality which are fitting seamlessly there, I suppose.

Issue #389, opened 2019-10-28
"send me reset password instructions" button not translated


vohmar:
The button to send password reset email is only in english

/sessions/password/new

![image](https://user-images.githubusercontent.com/7734747/67696769-1e55a380-f9b0-11e9-8f6e-8b1f98ae21fa.png)

vohmar:
et: "Saada parooli muutmise juhised"

vohmar:
fixed in #391

PR #391, opened 2019-10-29
390 password change translation

### LINKED ISSUE #390: Change password form partially translated
number of translations missing

image

yulgolem:
Closes #390 , #389

### DIFF
--- a/app/views/devise/passwords/edit.html.erb
+++ b/app/views/devise/passwords/edit.html.erb
@@ -23,7 +23,7 @@
                 <%= f.password_field :password_confirmation, autocomplete: "off" %>
             </div>
 
-            <%= f.submit t('.change_password'), class: "ui button primary" %>
+            <%= f.submit nil, class: 'ui button primary' %>
         <% end %>
 
         <%= render "devise/shared/links" %>
--- a/app/views/devise/passwords/edit.html.erb
+++ b/app/views/devise/passwords/edit.html.erb
@@ -11,19 +11,19 @@
             <%= f.hidden_field :reset_password_token %>
 
             <div class="field">
-                <%= f.label :password, "New password" %><br />
+                <%= f.label :password, t('.new_password') %><br />
                 <% if @minimum_password_length %>
-                    <em>(<%= @minimum_password_length %> characters minimum)</em>
+                    <em>(<%= t('.min_characters', min_chars: @minimum_password_length ) %>)</em>
                 <% end %>
                 <%= f.password_field :password, autofocus: true, autocomplete: "new-password" %>
             </div>
 
             <div class="field">
-                <%= f.label :password_confirmation, "Confirm new password" %><br />
+                <%= f.label :password_confirmation, t('.confirm_new_password') %><br />
                 <%= f.password_field :password_confirmation, autocomplete: "off" %>
             </div>
 
-            <%= f.submit "Change my password", class: "ui button primary" %>
+            <%= f.submit t('.change_password'), class: "ui button primary" %>
         <% end %>
 
         <%= render "devise/shared/links" %>
--- a/app/views/devise/passwords/new.html.erb
+++ b/app/views/devise/passwords/new.html.erb
@@ -13,8 +13,7 @@
                 <%= f.label :email %><br />
                 <%= f.email_field :email, autofocus: true, autocomplete: "email" %>
             </div>
-
-            <%= f.submit "Send me reset password instructions", class: "ui button primary" %>
+            <%= f.submit t('.reset_button'), class: "ui button primary" %>
         <% end %>
 
         <br>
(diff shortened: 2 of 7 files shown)


app/views/devise/passwords/new.html.erb
```diff
@@ -13,8 +13,7 @@
                 <%= f.label :email %><br />
                 <%= f.email_field :email, autofocus: true, autocomplete: "email" %>
             </div>
-
-            <%= f.submit "Send me reset password instructions", class: "ui button primary" %>
+            <%= f.submit t('.reset_button'), class: "ui button primary" %>
```
artur-beljajev: This name is super! It clearly expands the meaning of the translation key. It allows to change translations (say from `Reset` to `Send me reset password instructions` or any other) without changing the code. Also, opening translation file it's immediately clear how the key is used in UI (`button`, in this case).

app/views/devise/passwords/edit.html.erb
```diff
@@ -11,19 +11,19 @@
...
             </div>
 
             <div class="field">
-                <%= f.label :password_confirmation, "Confirm new password" %><br />
+                <%= f.label :password_confirmation, t('.confirm_new_password') %><br />
                 <%= f.password_field :password_confirmation, autocomplete: "off" %>
             </div>
 
-            <%= f.submit "Change my password", class: "ui button primary" %>
+            <%= f.submit t('.change_password'), class: "ui button primary" %>
```
artur-beljajev: Perhaps use built-in translations? (`submit` is i18n-aware method, as far as I remember).

config/locales/devise.et.yml
```diff
@@ -59,6 +59,13 @@ et:
       send_paranoid_instructions: "Kui sinu e-posti aadress on meie andmebaasis, saadetakse  mõne minuti jooksul sinu e-posti aadressile link salasõna taastamiseks."
       updated: "Salasõna on muudetud. Oled sisse logitud."
       updated_not_active: "Salasõna on muudetud."
+      new:
+        reset_button: "Saada parooli muutmise juhised"
+      edit:
+        new_password: "Uus parool"
+        min_characters: "Vähemalt %{min_chars} märki"
+        confirm_new_password: "Sisesta uus parool uuesti"
+        change_password: "Muuda mu parool"
```
artur-beljajev: Quotes are needed here?

test/system/users/forgotten_password_test.rb
```diff
@@ -28,4 +28,37 @@ def test_can_reset_password_by_email
...
+    assert(page.has_content?("Confirm new password"))
+    assert(page.has_button?("Change my password"))
+
+    click_link('Eesti keeles')
+
+    assert(page.has_content?("Uus parool"))
+    assert(page.has_content?("Vähemalt 8 märki"))
+    assert(page.has_content?("Sisesta uus parool uuesti"))
+    assert(page.has_button?("Muuda mu parool"))
+  end
```
artur-beljajev: This code does too much. I think it's safe to assume that if we put translations into YAML files, the UI will be translated. This is tested by Rails team. Otherwise such tests increase maintenance costs, take resources to run, with extremely little value.

config/locales/devise.et.yml
yulgolem: Not at all, just used them solely to support the overall translations style.

app/views/devise/passwords/edit.html.erb
yulgolem: Yep, but it will lead to moving the translations to global helpers i.e. en.helpers.submit.users.edit
Made  it so anyway.

test/system/users/forgotten_password_test.rb
yulgolem: Removed.

app/views/devise/passwords/edit.html.erb
```diff
@@ -11,19 +11,19 @@
...
             </div>
 
             <div class="field">
-                <%= f.label :password_confirmation, "Confirm new password" %><br />
+                <%= f.label :password_confirmation, t('.confirm_new_password') %><br />
                 <%= f.password_field :password_confirmation, autocomplete: "off" %>
             </div>
 
-            <%= f.submit "Change my password", class: "ui button primary" %>
+            <%= f.submit class: "ui button primary" %>
```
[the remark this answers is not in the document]
yulgolem: We can omit value here 'cause it taken from default I18n keys for model.
image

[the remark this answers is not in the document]
artur-beljajev: I am not sure it's a good idea since it's kinda unexpected without [this](https://github.com/rails/rails/blob/v5.0.7.2/actionview/lib/action_view/helpers/form_helper.rb#L1813). For me it seems like a hack and it encourages using it in that way.

```
def test(one = nil, two = {})
	p one
	p two
end
```

Result:
```
test(class: 'test')
p RUBY_VERSION
```

Notice that `class` was assigned to param `one`, not `two`.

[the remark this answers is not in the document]
yulgolem: Heh, I suppose it's more of semantic sugar than hack, and if it's implemented in framework it's legit AF... but I'll fix it anyway for the sake of consistency.

PR #393, opened 2019-10-29
Better test coverage for Estonia personal ID codes

karlerikounapuu:
At the moment, some of PR tests are failing because coverage for Estonia personal ID codes is poor. Provides better test coverage.

### DIFF
--- a/lib/identity_code/estonia.rb
+++ b/lib/identity_code/estonia.rb
@@ -37,7 +37,7 @@ class IdentityCode
 
       if checknum == 10
         scales = [3, 4, 5, 6, 7, 8, 9, 1, 2, 3]
-        checknum  = combine_scale_with_number(scales, identity_code)
+        checknum = combine_scale_with_number(scales, identity_code)
 
         checknum == 10 ? 0 : checknum
       else
--- a/lib/identity_code/estonia.rb
+++ b/lib/identity_code/estonia.rb
@@ -37,7 +37,7 @@ class IdentityCode
 
       if checknum == 10
         scales = [3, 4, 5, 6, 7, 8, 9, 1, 2, 3]
-        checknum = combine_scale_with_number(scales, identity_code)
+        checknum  = combine_scale_with_number(scales, identity_code)
 
         checknum == 10 ? 0 : checknum
       else
--- a/lib/identity_code/estonia.rb
+++ b/lib/identity_code/estonia.rb
@@ -4,6 +4,10 @@ class IdentityCode
       valid_length? && valid_century? && valid_birth_date? && valid_check_digit?
     end
 
+    def invalid?
+      !valid?
+    end
+
     private
 
     def valid_century?
--- a/test/lib/identity_code_test.rb
+++ b/test/lib/identity_code_test.rb
@@ -31,14 +31,29 @@ class IdentityCodeTest < ActiveSupport::TestCase
     assert(identity_code.valid?)
   end
 
-  def test_is_valid_for_every_century
+  def test_validates_estonian_id_codes_correctly
+
+    # Valid estonian ID code for person born in 19th century
     identity_code = IdentityCode.new('EE', "14501234213")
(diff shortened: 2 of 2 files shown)


lib/identity_code/estonia.rb
```diff
@@ -37,7 +37,7 @@ def check_digit
 
       if checknum == 10
         scales = [3, 4, 5, 6, 7, 8, 9, 1, 2, 3]
-        checknum  = combine_scale_with_number(scales, identity_code)
+        checknum = combine_scale_with_number(scales, identity_code)
```
artur-beljajev: Can we keep reformatting-related changes separately? Otherwise this change seems to have nothing to do with "Fix test coverage for Estonia personal ID codes".

test/lib/identity_code_test.rb
```diff
@@ -30,4 +30,15 @@ def test_is_valid_randomly_for_estonian_id_codes
     identity_code = IdentityCode.new('EE', array_of_codes.sample)
     assert(identity_code.valid?)
   end
+
+  def test_is_valid_for_every_century
```
artur-beljajev: - Looking at the name of the test, it's not clear what this test actually tests.
- I also find the tests themselves quite cryptic: why value `14501234213` is valid?
- Don't we need to test negative case (when a record is invalid)?

lib/identity_code/estonia.rb
karlerikounapuu: Alright. Opening a new PR just to fix one space feels hilarious to me, to be honest. I'll add the space back.

artur-beljajev: I meant separate commit, actually :)

artur-beljajev: Something that would explain the change, like "Reformat" (this is what I use, but feel free to choose whatever you think is clear for the widest audience)

artur-beljajev: The point is that if you see such commit message, you know right away that no logic is changed there, and you don't spend time digging into the details.

test/lib/identity_code_test.rb
karlerikounapuu: - Agreed. This test can be named more accurately.
- For me it's self-explanatory, I assert that this ID code is valid. Why would one write a test that intentionally is not valid? 
- Yup, great idea.

karlerikounapuu:
> > This PR pushes coverage to 100%.
> 
> Do you think this should be a goal?

Not a goal, but intention of this PR is to have better test coverage, yes. It was meant as a statement.

test/lib/identity_code_test.rb
```diff
@@ -30,4 +30,30 @@ def test_is_valid_randomly_for_estonian_id_codes
...
+    identity_code = IdentityCode.new('EE', "38701234218")
+    assert(identity_code.valid?)
+
+    # Valid estonian ID code for person born in 21th century
+    identity_code = IdentityCode.new('EE', "60901234211")
+    assert(identity_code.valid?)
+
+    # Incorrect estonian ID codes
+    identity_code = IdentityCode.new('EE', "29511234213")
+    assert_not(identity_code.valid?)
```
artur-beljajev: `assert identity_code.invalid?` ?

karlerikounapuu: identity_code doesn't have function `invalid?`. I wouldn't create this method because it doesn't solve anything nor provide functionality that's _really_ needed.

artur-beljajev: You're right! I assumed that it's `ActiveModel`-based for some reason... I would still create `invalid?` with the following simple content:

```
def invalid?
  !valid?
end
```

since this is what we're testing: that a record is invalid. Negations are less readable, so if it can be avoided, I would avoid it.

karlerikounapuu: Okay, fine by me 😀 Adding your snippet and porting tests to use it as well.

PR #394, opened 2019-11-01
343 health check

### LINKED ISSUE #343: Add health check status page
It's desired to have application based health status check functionality to make it easier for monitoring systems to get an overview of what works and what does not. If applicable this should be implemented to all our systems in unified manner, but let's start here

gem that may be of help here:
https://github.com/ianheggie/health_check

gitlab example:
https://docs.gitlab.com/ee/user/admin_area/monitoring/health_check.html

yulgolem:
Resolves #343 
Creates /healthcheck/all.json endpoint as well as /healthcheck/default.json, /healthcheck/database.json,
/healthcheck/api.json, /healthcheck/tara.json, /healthcheck/email.json, /healthcheck/sms.json, /healthcheck/registry.json

### DIFF
--- a/app/models/concerns/http_requester.rb
+++ b/app/models/concerns/http_requester.rb
@@ -1,0 +1,45 @@
+module Concerns
+  module HttpRequester
+    extend ActiveSupport::Concern
+
+    HTTP_METHODS = {
+      get: Net::HTTP::Get,
+      post: Net::HTTP::Post,
+    }.freeze
+
+    def default_request_response(url:, body:, headers:, type: :post)
+      uri = URI(url)
+      http = Net::HTTP.new(uri.host, uri.port)
+      http.use_ssl = (url.scheme == 'https')
+      request = HTTP_METHODS[type].new(uri.request_uri)
+      headers&.each { |key, val| request[key] = val }
+      request.body = body.to_json if body
+      success_result(response: http.request(request))
+    rescue Errno::ECONNREFUSED, Errno::EHOSTUNREACH, SocketError, Errno::EADDRNOTAVAIL
+      failed_result
+    end
+
+    def success_result(response:)
+      {
+        body: JSON.parse(response.read_body),
+        status: response.code.to_i,
+      }
+    end
+
+    def failed_result
+      error_code = Rack::Utils::SYMBOL_TO_STATUS_CODE[:service_unavailable]
+      {
+        body: 'Error occured',
+        status: error_code,
+      }
+    end
+
+    def default_post_request_response(url:, body: nil, headers: nil)
+      default_request_response(url: url, body: body, headers: headers, type: :post)
+    end
+
+    def default_get_request_response(url:, body: nil, headers: nil)
+      default_request_response(url: url, body: body, headers: headers, type: :get)
+    end
+  end
+end
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -1,5 +1,8 @@
 OkComputer.mount_at = false
-OkComputer.require_authentication(
-  ENV['HEALTHCHECK_USER'],
-  ENV['HEALTHCHECK_PASSWORD'],
-)
+OkComputer.check_in_parallel = true
+
+OkComputer::Registry.register 'api', ApplicationHealthCheck::API.new
+OkComputer::Registry.register 'registry', ApplicationHealthCheck::Registry.new
+OkComputer::Registry.register 'tara', ApplicationHealthCheck::Tara.new
+OkComputer::Registry.register 'email', ApplicationHealthCheck::Email.new
+OkComputer::Registry.register 'sms', ApplicationHealthCheck::SMS.new
--- a/lib/tasks/data_migrations/create_healthcheck_endpoints_settings.rake
+++ b/lib/tasks/data_migrations/create_healthcheck_endpoints_settings.rake
@@ -1,0 +1,37 @@
+namespace :data_migrations do
+  desc 'Seeds API, Tara & SMS healthcheck endpoints'
+  task create_healthcheck_endpoint_settings: :environment do
+
+    hash = {
+      check_api_url: {
+        description: <<~TEXT.squish,
+          URL to our own auction API endpoint for health checking.
+          Must be absolute, default is http://localhost/auctions.json.
+        TEXT
+        value: 'http://localhost/auctions.json',
+      },
+      check_sms_url: {
+        description: <<~TEXT.squish,
+          URL of SMS service provider for health checking.
+          Must be absolute.
+        TEXT
+        value: 'https://status.messente.com/api/v1/components/1',
+      },
+      check_tara_url: {
+        description: <<~TEXT.squish,
+          URL of OAUTH Tara provider for health checking.
+          Must be absolute.
+        TEXT
+        value: 'https://tara-test.ria.ee/oidc/jwks',
+      },
+    }
+
+    Setting.transaction do
+      hash.each do |key, value_hash|
+        setting = Setting.find_or_create_by(code: key)
+        setting.update!(value: value_hash[:value], description: value_hash[:description])
+      end
+      puts "Healthcheck API endpoints settings are set"
+    end
+  end
+end
--- a/test/integration/health_check_test.rb
+++ b/test/integration/health_check_test.rb
@@ -1,0 +1,71 @@
+require 'test_helper'
+
+class HealthCheckTest < ActionDispatch::IntegrationTest
+
+  def test_default_check_passed
+    get("/healthcheck/default", as: :json)
+
+    keys = parse_keys(response: response, endpoint: 'default')
+
+    assert_health_check_keys(keys)
+  end
+
+  def test_db_check_passed
+    get('/healthcheck/database', as: :json)
+
+    keys = parse_keys(response: response, endpoint: 'database')
+
+    assert_health_check_keys(keys)
+  end
+
+  def test_api_check_passed
+    get('/healthcheck/api', as: :json)
+
+    keys = parse_keys(response: response, endpoint: 'api')
+
+    assert_health_check_keys(keys)
+  end
+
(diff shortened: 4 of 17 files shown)


yulgolem:
> In general, when check is failing, error `500` is returned. It seemed to be confusing whether Health Check mechanism is failing or the thing being tested. I'm not sure what right / standard approach would be in this scenario, but for me returning `503` (service unavailable) seems to be more appropiate.

well, the error 500 is returned by HTTPRequestHelper, and then checker marks this check as failed. We would not receive 500 as a healthcheck error, only standard checker error response.

lib/http_request_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+    request.body = body.to_json if body
+    response = http.request(request)
+
+    {
+      body: JSON.parse(response.read_body),
+      status: response.code.to_i,
+    }
+  rescue
+    {body: 'Error occured',
+     status: 503,}
```
karlerikounapuu: Rubocop reports styling issues, spaces missing. Apart from that, looking good!

yulgolem: Hey, which version of rubocop.yml do you use? 
image

karlerikounapuu: Oops, my bad. I've modified `Documentation` to `Style/Documentation`

app/models/check/api.rb
```diff
@@ -0,0 +1,10 @@
+module Check
```
artur-beljajev: `Check` is too generic. How about `application_health_check`? Maybe put `Check` under `ApplicationHealth` module?

config/application.rb
```diff
@@ -28,6 +28,8 @@ class Application < Rails::Application
     # Load customization from special file
     config.customization = config_for(:customization)
 
+    config.autoload_paths << Rails.root.join('lib')
```
artur-beljajev: Avoid adding `lib` to `autoload_paths` or `eager_load_paths`. It loads everything from there, including (but not limited to) rake tasks, which is not desirable.

config/initializers/okcomputer.rb
```diff
@@ -1,5 +1,8 @@
 OkComputer.mount_at = false
-OkComputer.require_authentication(
-  ENV['HEALTHCHECK_USER'],
-  ENV['HEALTHCHECK_PASSWORD'],
-)
+OkComputer.check_in_parallel = true
+
+OkComputer::Registry.register "api", Check::API.new
```
artur-beljajev: Single quotes?

lib/http_request_helper.rb
```diff
@@ -0,0 +1,32 @@
+module HTTPRequestHelper
```
artur-beljajev: What is the purpose of this module? Can it be implemented as a normal class?

lib/tasks/data_migrations/create_healthcheck_endpoints_settings.rake
```diff
@@ -0,0 +1,37 @@
...
+  desc 'Seeds API, Tara & SMS healthcheck endpoints'
+  task create_healthcheck_endpoint_settings: :environment do
+
+    hash = {
+      check_api_url: {
+        description: <<~TEXT.squish,
+          URL to our own auction API endpoint for health checking.
+          Must be absolute, default is http://lvh.me:3000/auctions.json.
+        TEXT
+        value: 'http://lvh.me:3000/auctions.json',
```
artur-beljajev: Why not just `localhost? Or what's the point of using `lhv.me` in this case? Is it really on port 3000?

app/models/check/base.rb
```diff
@@ -0,0 +1,25 @@
+module Check
+  class Base < OkComputer::Check
```
artur-beljajev: Consider refactoring. To me it looks like case no. 2 from https://stackoverflow.com/a/2947823/2987689. If `Base` is an abstract class, testing it looks a bit weird. It looks like there are some responsibilities that do not belong to this class.

app/models/check/base.rb
```diff
@@ -0,0 +1,25 @@
...
+    def headers
+      { 'Content-Type': 'application/json' }
+    end
+
+    def simple_check_endpoint(url:, no_url_message:, fail_message:, success_message:)
+      if url
+        url = URI(url)
+        result = default_get_request_response(url: url, headers: headers)
+
+        result[:status] == 200 ? mark_message(success_message) : report_failure(fail_message)
```
artur-beljajev: Magic number. Check https://gist.github.com/mlanett/a31c340b132ddefa9cca#gistcomment-2184336 and https://gist.github.com/mlanett/a31c340b132ddefa9cca#gistcomment-2988870.

app/models/check/email.rb
```diff
@@ -0,0 +1,40 @@
...
+      if host.present? && port.present?
+        check_smtp(host, port.to_i)
+      else
+        report_failure('Mail server host or port not set')
+      end
+    end
+
+    private
+
+    def check_smtp(host, port)
```
artur-beljajev: Keyword arguments to avoid https://connascence.io/position.html?

app/models/check/email.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    def host
+      Rails.application.config.action_mailer.smtp_settings[:address]
+    end
+
+    def port
+      Rails.application.config.action_mailer.smtp_settings[:port]
+    end
+
+    def port_open?(ip, port, seconds = 1)
```
artur-beljajev: Keyword arguments. Don't get why we rescue exceptions.

yulgolem:
> Why JSON is tested in system tests using Capybara API? They seem like integration ones.

Well, it's not integration per se, as we are testing part of system- healthcheck endpoint. This endpoint is integrated with smtp server/sms service provider/etc, but we are to check if our endpoint online and well, not integration part.

lib/http_request_helper.rb
yulgolem: This module totally is a helper for simplifying http-requests. Moved it to /helpers/ anyway.

lib/tasks/data_migrations/create_healthcheck_endpoints_settings.rake
yulgolem: Shame on me. 'll move it to just localhost.

app/models/check/base.rb
yulgolem: Moved to helper.

app/models/check/email.rb
yulgolem: Rescues here means what we are unable to connect to port by some reasons - timeout or some network or address errors. We are not interested in the exact reason we are cannot connect, but in the exact fact - this exact port on this address in unreachable.

yulgolem:
@artur-beljajev ready to check.

app/models/concerns/http_requester.rb
```diff
@@ -0,0 +1,44 @@
...
+      {
+        body: JSON.parse(response.read_body),
+        status: response.code.to_i,
+      }
+    end
+
+    def failed_result
+      {
+        body: 'Error occured',
+        status: 503,
```
artur-beljajev: What does this number mean?

yulgolem: Changed to symbolic :service_unavailable.

yulgolem:
> https://github.com/internetee/auction_center/pull/394/files#diff-3f95dea6b31e219d225915bc7c332f86 should be rethought. Testing JSON using Capybara and running it through headless chrome looks strange.

Moved it to integration tests, testing JSON only now.

test/integration/health_check_test.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  def test_email_check_passed
+    endpoint = 'email'
+    get("/healthcheck/#{endpoint}.json", headers: @headers)
+    keys = parse_keys(response: response, endpoint: endpoint)
+
+    assert_health_check_keys(keys)
+  end
+
+  def test_sms_check_passed
```
artur-beljajev: What do you think about grouping the statements according to `given/when/then` steps visually?

artur-beljajev: Like this:
 
```
# Given. absent?

# When.
get("/healthcheck/sms.json", headers: @headers)

# Then.
keys = parse_keys(response: response, endpoint: endpoint)
assert_health_check_keys(keys)
```

artur-beljajev: P.S. Comments are just to explain my point. No need to keep them.

artur-beljajev: Currently it's not clear at first sight, where is what.

test/models/application_health_check/sms_test.rb
```diff
@@ -0,0 +1,11 @@
+require 'test_helper'
+
+class SMSTest < ActiveSupport::TestCase
+
```
artur-beljajev: Remove this.

test/integration/health_check_test.rb
yulgolem: I suppose here we've got given: endpoint.

test/integration/health_check_test.rb
```diff
@@ -0,0 +1,76 @@
...
+
+class HealthCheckTest < ActionDispatch::IntegrationTest
+
+  def setup
+    super
+    @headers = { 'Accept': 'application/json' }
+  end
+
+  def test_default_check_passed
+    get("/healthcheck/default.json", headers: @headers)
```
artur-beljajev: - I think it should be `get /healthcheck/default`, headers: ..., as: :json`.
- Wouldn't it be better to use name routes here rather than hardcoding?

yulgolem: Unfortunately, this is a OkComputer::Engine inner routes, the only way to get exact route here is to have a hardcoded URL, other way (if we use something like :healthcheck_url) we've get only one default endpoint for overall check.
But changed to as: :json anyway.

artur-beljajev: seems perfect!

PR #399, opened 2019-11-04
Prevent user from deleting billing profile with active offers

### LINKED ISSUE #313: Possible to delete billing profile with active offers
User is able to delete the billing profile that they just made an offer with. As a result auction system is not able to generate invoice in case the offer is the best. It should not be possible to delete billing profile that has active offer associated to it - user should first change the profile in the offer.

Also if user tries to delete their profile associated with an invoice it fails with no message. There should be an error message stating the reason why operation failed.

karlerikounapuu:
Closes #313 

- User receives error message when trying to delete Billing Profile which is used for active offer
- BillingProfile address is mirrored to Invoice. BillingProfile doesn't need to "belong" to Invoice anymore.
- Static billing address for paid Invoice. Eg address is not mirrored to Invoice if it's already paid.


app/controllers/admin/results_controller.rb
```diff
@@ -47,5 +51,11 @@ def search_params
       search_params_copy = params.dup
       search_params_copy.permit(:domain_name)
     end
+
+    def buyer_name
+      return @result.offer.billing_profile.name if @result.offer.billing_profile.present?
+
+      Invoice.find_by(result: @result.offer).recipient
```
yulgolem: We are pretty sure that this invoice always exists? Just wondering.

yulgolem: OK got it.

app/controllers/billing_profiles_controller.rb
```diff
@@ -52,10 +52,10 @@ def update
   # DELETE /billing_profiles/aa450f1a-45e2-4f22-b2c3-f5f46b5f906b
   def destroy
     @billing_profile.destroy
-
-    respond_to do |format|
-      format.html { redirect_to billing_profiles_path, notice: t(:deleted) }
-      format.json { head :no_content }
+    if @billing_profile.destroyed?
```
artur-beljajev: https://github.com/internetee/auction_center/commit/29678320460e2f9972edf8e6bae2402dd451c86f

Quite similar case. Consider using `destroy!`. If there is some description involved (about the reason of being unable to delete), consider moving it to UI. If something cannot be deleted, it may worth to restrict this action and give some hint to a user in UI.

test/system/billing_profiles_test.rb
```diff
@@ -151,6 +152,16 @@ def test_a_user_can_delete_their_billing_profile
     assert_text('Deleted successfully.')
   end
 
+  def test_a_user_can_not_delete_profile_in_use
```
artur-beljajev: See comment above.

test/models/billing_profile_test.rb
```diff
@@ -102,4 +102,13 @@ def test_create_default_for_user_creates_a_billing_profile
     assert_equal(billing_profile.country_code, user_without_billing_profile.country_code)
     assert_equal(billing_profile.name, user_without_billing_profile.display_name)
   end
+
+  def test_billing_address_is_mirrored_to_linked_invoice
+    # Invoice which is linked with @billing_profile
```
artur-beljajev: How about embodying this knowledge in the code? like `invoice_with_billing_profile`, which becomes a private method of `BillingProfileTest`?

test/models/billing_profile_test.rb
```diff
@@ -102,4 +102,13 @@ def test_create_default_for_user_creates_a_billing_profile
...
+    invoice = invoices(:payable)
+    @billing_profile.update(name: "New Company Ltd")
```
artur-beljajev: - Single quotes.
- Can we extract this literal into variable `billing_profile_name` (this name sounds weird to me, in fact, but this is a design problem, and out of this PR's context) and reuse it?

config/locales/et.yml
```diff
@@ -11,6 +11,7 @@ et:
   created: "Loodud."
   updated: "Uuendatud."
   deleted: "Kustutatud."
+  something_went_wrong: "Midagi läks valesti."
```
artur-beljajev: The intention is to translate exiting key from `en` locale? Or how is it connected to this PR?

app/models/invoice.rb
```diff
@@ -2,13 +2,15 @@
 require 'result_not_sold'
 
 class Invoice < ApplicationRecord
+  include BillingHelpers
```
artur-beljajev: I would avoid `Helper` word. It doesn't explain the role an object plays. Consider refactoring. Maybe extract `address` and `vat_rate` into their own objects. As a quick fix they might remain right in `Invoice`. Otherwise by looking at `Invoice` class I have no idea what that `BillingHelpers` includes.

app/models/concerns/billing_helpers.rb
```diff
@@ -0,0 +1,20 @@
...
+  extend ActiveSupport::Concern
+
+  def address
+    country_name = Countries.name_from_alpha2_code(country_code)
+    postal_code_with_city = [postal_code, city].join(' ')
+    [street, postal_code_with_city, country_name].compact.join(', ')
+  end
+
+  def vat_rate
+    if country_code == 'EE'
```
artur-beljajev: Can it be configurable? This prevents other users of software besides us from using it, since country will differ, but logic stays the same.

app/models/billing_profile.rb
```diff
@@ -44,4 +30,31 @@ def self.create_default_for_user(user_id)
...
   end
+
+  def check_no_active_offers_associated
+    return unless active_offers?
+
+    errors.add(:base, I18n.t('billing_profiles.in_use_by_offer'))
+    throw :abort
+  end
+
+  def active_offers?
```
artur-beljajev: This seems quite complicated. Consider refactoring (extract methods).

app/models/billing_profile.rb
```diff
@@ -1,19 +1,16 @@
...
   validates :vat_code, uniqueness: { scope: :user_id }, allow_blank: true
 
   belongs_to :user, optional: true
-
-  def address
-    country_name = Countries.name_from_alpha2_code(country_code)
-    postal_code_with_city = [postal_code, city].join(' ')
-    [street, postal_code_with_city, state, country_name].compact.join(', ')
-  end
+  before_destroy :check_no_active_offers_associated
```
artur-beljajev: Any chance to avoid callbacks? It comes difficult to maintain such code. Perhaps introducing `deletable?` predicate can help?

app/models/offer.rb
```diff
@@ -41,6 +41,10 @@ def price=(value)
   end
 
   def total
-    price * (1 + billing_profile.vat_rate)
+    if billing_profile.present?
+      price * (1 + billing_profile.vat_rate)
```
artur-beljajev: Maybe extract method can help to better explain this magic number?

db/migrate/20191022125038_add_billing_address_to_invoice.rb
```diff
@@ -0,0 +1,12 @@
+class AddBillingAddressToInvoice < ActiveRecord::Migration[5.2]
+  def change
+    add_column :invoices, :recipient, :string
+    add_column :invoices, :vat_code, :string
```
artur-beljajev: `vat_number`, not code.

app/models/offer.rb
yulgolem: Or just a freezed named constant value?

lib/tasks/data_migrations/invoice.rake
```diff
@@ -0,0 +1,32 @@
...
+  task populate_billing_fields: :environment do
+    @invoices = Invoice.all
+    migrated_invoice_count = 0
+    failed_invoices = []
+
+    @invoices.each do |invoice|
+      next if invoice.billing_profile.blank?
+      next unless invoice.recipient.nil?
+
+      fields = %w[vat_code street city state postal_code alpha_two_country_code]
```
yulgolem: We've already got this [here](https://github.com/internetee/auction_center/pull/399/files#diff-fcf17760e7a27b92c25b76c075e48b86R52). How about to just reuse this method?

Issue #212, opened 2019-04-05
Update Ruby to 2.6


maciej-szlosarczyk:
We have been running tests against 2.6 since March 2019, and they have been rather successful.

Requires changes to docker images.

PR #404, opened 2019-11-06
Upgrade Ruby to 2.6.5, Puma => 4.2.1

karlerikounapuu:
- Upgraded Ruby to 2.6.5
- Updated gems conservatively
- Dropped Ruby 2.5 checks in Travis
- Upgraded Puma to 4.2.1. Resolves an issue with broken reactor loop while running tests.


vohmar:
closes #212 
@karlerikounapuu Please take a look at Docker images - some changes required there as well

artur-beljajev:
@karlerikounapuu Can we have separate PR for Ruby upgrade? I find a good practice to have separate commit for every gem update.

It seems there might be one single commit upgrading Ruby. Currently if we need to revert Ruby upgrade, we need to revert two commits., which actually represent single working unit.

Gemfile
```diff
@@ -17,11 +17,12 @@ gem 'omniauth-rails_csrf_protection'
 gem 'omniauth-tara', github: 'internetee/omniauth-tara'
 gem 'pdfkit'
 gem 'pg', '>= 0.18', '< 2.0'
-gem 'puma', '~> 3.11'
+gem 'puma', '~> 4.2.1'
 gem 'rails', '~> 5.2.1'
 gem 'rails-i18n'
 gem 'recaptcha'
 gem 'simpleidn'
+gem 'sprockets', '~> 3.7.2'
```
artur-beljajev: Why you lock it to `patch` version?

artur-beljajev:
@karlerikounapuu Just to be sure: I guess you checked Ruby 2.6 changelog to be sure our code (and possible 3rd dependencies' code) is compatible with it?

artur-beljajev:
@karlerikounapuu By the way, is Rails 5.2 fully compatible with Ruby 2.6?

karlerikounapuu:
> @karlerikounapuu Can we have separate PR for Ruby upgrade? Also I find a good practice to have separate commit for every gem update.
> 
> It seems there might be one single commit upgrading Ruby. Currently if we need to revert Ruby upgrade, we need to revert two commits., which actually represent single working unit.

Yes, might be a great idea. I'm opening a new PR regarding to plugin updates and keeping Ruby version related stuff here.

artur-beljajev:
Changed PR title from "Upgrade Ruby to 2.6.5, Puma to 4.2.1", to "Upgrade Ruby to 2.6.5", if you don't mind.

karlerikounapuu:
> I changed PR title from "Upgrade Ruby to 2.6.5, Puma to 4.2.1", to "Upgrade Ruby to 2.6.5", if you don't mind. :)
> 
> Oh, I didn't notice that there is no review notification pending. Is it ready for review at all?
@artur-beljajev 
Forgot about changing name for PR, great.

Yup, this one is ready for review. I'm opening gem upgrade PR soon, but it's not really dependent on this PR I guess.

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: ruby
 rvm:
-  - 2.5
-  - 2.6
+  - 2.6.5
```
artur-beljajev: Can we drop `patch` version? I am not sure it's worth the effort update it every time a new `patch` version is released. There can be no backwards-incompatible changes in both `minor` or `patch` version according to semver.

karlerikounapuu: Sure can do, no problem.

artur-beljajev:
Given 
> Bundler is now installed as a default gem.

Do we still need `gem install bundler` in `Dockerfile`s? https://github.com/internetee/auction_center/blob/f736d21d68d5b82e621d7f2cef728183ee264fd0/Dockerfile#L6?

https://www.ruby-lang.org/en/news/2018/12/25/ruby-2-6-0-released/

karlerikounapuu:
> Given
> 
> > Bundler is now installed as a default gem.
> 
> Do we still need `gem install bundler` in `Dockerfile`s?
> 
> https://github.com/internetee/auction_center/blob/f736d21d68d5b82e621d7f2cef728183ee264fd0/Dockerfile#L6
> 
> ?
> https://www.ruby-lang.org/en/news/2018/12/25/ruby-2-6-0-released/

That's something I've done out of habit honestly. Used to run into an issue when older bundler (1.17.2) was shipped with Ruby whereas Gemfile was created using 2.0.2. I'll look into it if they've updated it. I'd like to keep them at same version.

artur-beljajev:
OK, let it be, not a big deal.

PR #405, opened 2019-11-07
Change confirmation message if user changes email

### LINKED ISSUE #402: Improve notification on email address change
E-mail address change requires user to verify access to the new address. For this confirmation email is sent and user profile is updated only after new address confirmation.

The message that user receives in the UI on email address update is just

Updated successfully

The message is misleading. The message should say

en: Confirmation link was sent to new email address. Please confirm the address for the change to take an effect!

et: Palun kinnita eposti aadressi muutus klõpsates uuele aadressile saadetud ekirjas olevale kinnituse lingile.

yulgolem:
Closes #402 , adds localised notification if email was changed.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -48,9 +48,16 @@ class UsersController < ApplicationController
   # PUT /users/aa450f1a-45e2-4f22-b2c3-f5f46b5f906b
   def update
     if valid_password?
+      @user.attributes = params_for_update
+      email_changed = @user.email_changed?
+
       respond_to do |format|
-        if @user.update(update_params)
-          format.html { redirect_to user_path(@user.uuid), notice: t(:updated) }
+        if @user.valid?
+          @user.save!
+
+          format.html do
+            redirect_to user_path(@user.uuid), notice: notification_for_update(email_changed)
+          end
           format.json { render :show, status: :ok, location: @user }
         else
           format.html { render :edit }
@@ -79,7 +86,7 @@ class UsersController < ApplicationController
                   :given_names, :surname, :mobile_phone, :accepts_terms_and_conditions, :locale)
   end
 
-  def update_params
+  def params_for_update
     update_params = params.require(:user)
                           .permit(:email, :password, :password_confirmation, :country_code,
                                   :given_names, :surname, :mobile_phone,
@@ -103,5 +110,9 @@ class UsersController < ApplicationController
 
   def authorize_user
     authorize! :manage, @user
+  end
+
+  def notification_for_update(email_changed)
+    email_changed ? t('.email_changed') : t(:updated)
   end
 end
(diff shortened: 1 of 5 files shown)


app/controllers/users_controller.rb
```diff
@@ -104,4 +104,9 @@ def set_user
   def authorize_user
     authorize! :manage, @user
   end
+
+  def update_notice
+    email_changed = @user.email != update_params[:email]
```
artur-beljajev: Did you consider `user.email_changed?` from Rails?

test/system/users/edit_user_test.rb
```diff
@@ -27,7 +27,7 @@ def test_update_of_email_requires_confirmation
     fill_in('user[email]', with: 'updated-email@auction.test')
     fill_in('user[current_password]', with: 'password123')
     click_link_or_button('Update')
-    assert(page.has_css?('div.notice', text: 'Updated successfully.'))
+    assert(page.has_css?('div.notice', text: 'Confirmation link was sent to new email address. Please confirm the address for the change to take an effect!'))
```
artur-beljajev: Do we need to rely on HTML structure here? What if we just test it from pure user's perspective?

Like `assert_text 'Confirmation link was sent to new email address. Please confirm the address for the change to take an effect!'`?

app/controllers/users_controller.rb
yulgolem: Yep, and here is the pitfall - we are not assigning attributes here and thus email attribute is not changed.
And we are updating via #update, not saving, and I totally do not want to make a callback in model on update.

test/system/users/edit_user_test.rb
yulgolem: Great idea, I'll do that.

app/controllers/users_controller.rb
artur-beljajev: What do you think about https://github.com/internetee/auction_center/pull/405/commits/0bc3824309ca92dd4338196d1c80c7fc854fba2b?

artur-beljajev: I made a mistake, updated.

yulgolem: I think that's excellent clarification allowing us to use dirty attributes. Thanks a lot!

artur-beljajev:
Thoughts/comments?

https://github.com/internetee/auction_center/pull/405/commits/29678320460e2f9972edf8e6bae2402dd451c86f

yulgolem:
> @yulgolem Thoughts/comments?
> 
> [2967832](https://github.com/internetee/auction_center/commit/29678320460e2f9972edf8e6bae2402dd451c86f) & [a01f5c2](https://github.com/internetee/auction_center/commit/a01f5c278239a087d90dfcf94be598ea9b778d31).

Well, I agree with that. 
I totally dislike to use any callbacks on models, however it can be necessary in some cases, and Active Record surely can silently fail for reasons other than validation errors - AR callbacks that do not return true (eg before_save), invalid parent/ child records etc.

Issue #400, opened 2019-11-05
Admin: inconsistent sorting behavior


vohmar:
Related to: #321 

Similarly to auctions view other admin views suffer from some unexpected sorting behavior as well

* /admin/results
  * "doman name", "finished at" and "registration due date" override sorting selected on other columns - status, name. Example:
 set ascending order on column 'status':
![image](https://user-images.githubusercontent.com/7734747/68199114-0db3b780-ffc6-11e9-86ef-39671823c7c4.png)

  * now trying to sort the resulting list by 'registration due date' in descending order
![image](https://user-images.githubusercontent.com/7734747/68199277-671be680-ffc6-11e9-8861-cbe7e452eeb3.png)

  * while the other way around. Reset the list and sorted domains by 'finished at' desc
![image](https://user-images.githubusercontent.com/7734747/68199534-e1e50180-ffc6-11e9-9a18-3d70b8143720.png)

  * and now sorting the results by 'name' asc the pervious sorting order is preserved
![image](https://user-images.githubusercontent.com/7734747/68200751-296c8d00-ffc9-11e9-9693-b0b062ac3df7.png)

* /admin/auctions
  * sililarly to results view - 'domain name' and 'ends at' columns override periously set order while the other columns do not do that.

* /admin/billing_profiles
  * column 'name' overrides all previous sorting 

* /admin/users
  * column 'email' overrides previously set sorting
  * sorting option on column 'roles' is missing

* /admin/bans
  * nothing overrides  'domain name' and 'invoice' and these columns override everything except each other
  * everything overrides 'user'
  * 'valid until' is overriden by 'valid from' in addition to 'domain name' and 'invoice'

* /admin/invoices
  * column 'billing profile' overrides all other column orders
  * columns 'due date' and 'status' override all but 'billing profile' sorting orders
  * sorting by currency columns does not work - 'total', 'total paid'

yulgolem:
The main issue here is following:
On sorting we are totally dependent on order of parameters in URL. We are assuming what then we are sorting by one column and this column is added un URL params, the second sorting column will be added after the first... and we are wrong.

The URL generation uses Hash#to_params method, which sorts parameters alphabetically. This is a known Rails feature [since 2011](https://github.com/rails/rails/issues/1146). So - then we are adding second, third etc sorting parameter, they are sorted in URL (and in dataset query) not by time of adding, but by their alphabetic order.

The solution is to sort datasets only by one column at a time, removing previous sorting params from URL.

[...]

PR #408, opened 2019-11-11
Simplify sorting behavior in admin tables

yulgolem:
Fixing #400 by implementing sorting by one column per time.

### DIFF
--- a/test/system/admin/admin_orderable_test.rb
+++ b/test/system/admin/admin_orderable_test.rb
@@ -52,6 +52,15 @@ class AdminOrderableTest < ApplicationSystemTestCase
     assert_appears_before('Orphan Profile', 'ACME Inc.')
   end
 
+  def test_order_links_are_with_single_parameter
+    visit admin_auctions_path
+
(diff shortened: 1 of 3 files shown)


test/system/admin/admin_orderable_test.rb
```diff
@@ -52,6 +52,15 @@ def test_invoices_are_orderable
     assert_appears_before('Orphan Profile', 'ACME Inc.')
   end
 
+  def test_order_links_are_with_single_parameter
+    visit admin_auctions_path
+
+    sorting_href = '/admin/auctions?order%5Badmin_auction_decorators.highest_offer_cents%5D=desc'
```
artur-beljajev: - Why hardcoded URL is needed here? Can we use named routes?
- What does this test verify? It is not clear why that `sorting_href` is expected to be present on the page.

yulgolem: This test if verify what there is only one sorting parameter in the button URL after we set sorting. Previous behavior was adding sorting params per each button, now it's only one param per button and that's all.

PR #416, opened 2019-11-15
Bump gem versions

karlerikounapuu:
Second part of upgrading Ruby to 2.6.5

- Bump Puma to 4.3.0. Resolves an issue with reactor error loops.
- Bump airbrake to 9.5.0
- Bump bootsnap to 1.4.5
- Bump money to 6.13.4
- Bump okcomputer to 1.18.0
- Bump recaptcha to 5.2.1
- Bump turbolinks to 5.2.1
- Bump webpacker to 4.2.0
- Bump wkhtmltopdf-binary to 0.12.5
- Bump capybara to 3.29.0
- Bump simplecov to 0.17.1
- Lock Sprockets to ~> 3.7. Sprockets 4.x has a new approach on how to deliver assets, but we mainly use Webpacker so there's no reason to migrate to new version.


artur-beljajev:
> Lock Sprockets to ~> 3.7. Sprockets 4.x has a new approach on how to deliver assets, but we mainly use Webpacker so there's no reason to migrate to new version.

It's worth putting this description into the code. Otherwise it will be a mystery why it is locked just to this version when looking at Gemfile.

PR #418, opened 2019-11-18
174 internal statistics

### LINKED ISSUE #174: Admin: statistics
Add view for statistics in admin interface

* number of auctions vs domains with and without bid - per day (graph)
* average number of bids per domain that had at least 1 bid - per day (graph)
* average number of bids - per day, per month, per year (graph)

* metric of paid but unregistered domains - count with configurable time gap (slider?)
* ratio of paid/updaid invoices (percentage) - graph
* ratio of registered/unregistered domains vs number of invoices (percentage) - graph
* ratio of registered/unregistered domains vs paid invoices (percentage) - graph

* auction winners - private vs organisations - bar chart with split bars indiationg ratio
* geography of auction winners (total)- map
* auction wins per user (total) - pie chart

yulgolem:
Closes #174 by adding internal statistics page


app/models/statistics_report.rb
```diff
@@ -0,0 +1,30 @@
+class StatisticsReport
+  attr_reader :start_date
+  attr_reader :end_date
+
+  REPORTS_KLASSES = [StatisticsReport::AuctionData,
```
karlerikounapuu: `REPORT_CLASSES`?

app/models/statistics_report/domains_data.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    def gather_data
+      (start_date..end_date).each do |date|
+        unregistered_domains_count_by_end[date] = unregistered_domains[date]&.count.to_i
+        registered_domains_count_by_end[date] = registered_domains[date]&.count.to_i
+      end
+    end
+
+    def unregistered_domains
+      Result.unregistered
```
karlerikounapuu: Wouldn't it be better to move this query to Result model scope / method?

yulgolem:
> When statistics page is accessed via menu link, chartkick doesn't load but stays in "loading" state. Works fine when page is accessed by URL / refreshed.

> ...and that happens because of turbolinks. Easiest (but not ideal) method to fix the issue would be to append `data: { turbolinks: false }` to link which redirects to statistics page.

Thanks a lot, fixed this.

app/models/statistics_report/invoice_data.rb
```diff
@@ -0,0 +1,58 @@
...
+    attr_accessor(*ATTRS)
+
+    def initialize(start_date:, end_date:)
+      @start_date = start_date
+      @end_date = end_date
+      ATTRS.each { |attr| send("#{attr}=", {}) }
+    end
+
+    def gather_data
+      invoices_by_date = Invoice.where(issue_date: start_date..end_date).group_by(&:issue_date)
```
karlerikounapuu: Maybe use Invoice scope as well?

app/models/statistics_report.rb
yulgolem: That's intended as a habit not to use keyword 'class' and use 'klass' instead.

app/models/statistics_report/domains_data.rb
yulgolem: I prefer not to clog model scopes with anything what used just in one place.

yulgolem:
Lacking the view refresh job.

PR #423, opened 2019-11-21
Fetch footer from main internet.ee site

### LINKED ISSUE #203: Shared page footer and main menu elements from Voog
To provide unified user experience over all of EIS's web interfaces we try to use same page footer and similar main menu structure on all the environments. To make this happen and avoid duplicate work it is possible to get the neccessary infromation from voog using APIs (there might be a need for webdevs to make our current page compatible with this, not sure). Reply form voog support:

> Tere!
> 
> Vool on üsna võimekas API, kust saab masinloetaval kujul nii mõndagi pärida. See, kuidas saidi andmed mallis välja näidatakse, on küll puhtalt malli arendaja / disaineri otsus. Aga näiteks peamenüü koostatakse enamasti saidi struktuurist - selle saab kätte läbi Node API [1]. "Node" on siis struktuuripuu tipp, mille küljes võib olla mitu (erinevas keeles) lehekülge (või elementi või blogiartiklit). Keerulisemaid andmestruktuure hoitakse Voos enamasti kataloogides (vt. Sisu → Kataloogid); sellele vastab Elements API [2]. Kui soovite ligi pääseda otse leheküljel olevatele sisualadele, siis tuleks kasutada Contents API-t [3]. Ehk siis suur osa Voo saidi sisust on masinloetav ja -kirjutatav. Soovitan uurida API dokumentatsiooni või küsida täpsemalt juba Voo arendajatele mõeldud Slacki-kanalis [4].
> 
> [1] https://www.voog.com/developers/api/resources/nodes
> [2] https://www.voog.com/developers/api/resources/elements
> [3] https://www.voog.com/developers/api/resources/contents
> [4] https://slack.voog.com/
> 
> Kõike head,

karlerikounapuu:
Closes #203 

Requires `rails data_migrations:create_shared_footer_fetcher_settings` to create nessecary settings. Otherwise UI is not rendered.


app/models/setting.rb
```diff
@@ -5,6 +5,9 @@ class Setting < ApplicationRecord
   validates :value, presence: true
   validates :code, uniqueness: true
 
+  scope :voog_site_url, -> { find_by(code: :voog_site_url)&.value }
```
yulgolem: I suppose better use class methods here.
You see, main scopes usage is to return `ActiveRecord::Relation`, not an instance of class, and scopes are intended to be linked (chained) to other relations. And here you return not only instance, but an attribute of this instance, it's not so good behaviour preventing us to chain this scope like `Auction.active.last` and so.

test/jobs/shared_footer_fetcher_job_test.rb
```diff
@@ -0,0 +1,37 @@
...
+  def setup
+    super
+    @voog_site_url = settings(:application_name)
+    @voog_api_key = settings(:auction_currency)
+    @voog_site_fetching_enabled = settings(:auction_minimum_offer)
+
+    @voog_site_url.update!(code: 'voog_site_url', value: 'https://test.url')
+    @voog_api_key.update!(code: 'voog_api_key', value: '123')
+    @voog_site_fetching_enabled.update!(code: 'voog_site_fetching_enabled', value: 'true')
+
```
yulgolem: Do we really want empty line here?

karlerikounapuu: Can be omitted. Thanks!

app/models/setting.rb
karlerikounapuu: Makes sense, yup. Making changes.

test/test_helper.rb
```diff
@@ -8,8 +8,10 @@
 require_relative '../config/environment'
 require 'rails/test_help'
 require 'minitest/mock'
+require 'webmock/minitest'
 
 class ActiveSupport::TestCase
+  WebMock.allow_net_connect!
```
artur-beljajev: Doesn't this make `webmock` gem useless? Perhaps it's better to allow some concrete URL on a per-test basis? I wonder what you want to archive by this, still.

karlerikounapuu: Actually it doesn't, I did this intentionally. I didn't want webmock to interfere with other tests since it's only being used in `SharedFooterFetcherJobTest` and with current setup stubs are working great. But if you see this as a concern, I could take a look again, sure.

lib/tasks/data_migrations/create_shared_footer_properties.rb.rake
```diff
@@ -0,0 +1,33 @@
+namespace :data_migrations do
+  desc 'Creates settings to fetch shared footer from VOOG site'
+  task create_shared_footer_fetcher_settings: :environment do
+    hash = {
+      voog_site_url: {
+        description: <<~TEXT.squish,
+          VOOG site from which to fetch localized footer elements. Defaults to https://www.internet.ee.
+        TEXT
+        value: 'https://www.internet.ee',
```
yulgolem: Maybe we should use the new Setting structure with formats?

karlerikounapuu: Totally missed this one. Thanks!

Issue #429, opened 2019-11-26
Health check API documentation 


vohmar:
We are missing documentation on health check API - specifically how to communicate with the API - which checks and what kind of statuses  are available.

Related to #343

yulgolem:
Application healthcheck API is available at the `/healthcheck` base endpoint and responds in plain text and json formats. There is no auth on this endpoint at the moment, but BasicAuth support can be plugged in if needed.

Supported checks are following:

- Default check, checks if application is running. Text output via `healthcheck/default`, JSON via `healthcheck/default.json`.  JSON Example:
`{"default":{"message":"Application is running","success":true,"time":2.00001522898674e-06}}`

- DB check, checks if there is a ActiveRecord database connection. Text output via `healthcheck/database`, JSON via `healthcheck/database.json`. JSON example:
`{"database":{"message":"Schema version: 20191025092912","success":true,"time":0.005994000006467104}}`
Sets JSON 'success' field to false if DB connection is down.

- Internal API check, checks if GET-request to the internal API url (which is set by Setting key `check_api_url`) returns HTTP-status 200 OK. Text output via `healthcheck/api`, JSON output via `healthcheck/api.json`. Sets JSON 'success' field to false if there is no API url set or if API url responds with anything other than 200 OK. See JSON 'message' field for error description.

- All checks in one page. Text output via endpoint `/healthcheck/all` , JSON output via `/healthcheck/all.json`.

vohmar:
please add the documentation to Readme (https://github.com/internetee/auction_center/blob/master/README.md) in the API subsection. Split this paragraph in two Auction API and Health Check API. The description can be right there or linked to separate doc page.

PR #430, opened 2019-11-26
Fixes response bug if registry integration disabled

yulgolem:
Fixes a bug found on #429 - if registry integration was disabled, check endpoint was throwing wrong error message.


yulgolem:
> I'm wondering if it's somehow possible to test highlighted line also? If not, that's fine as well.

I'd omit it for now - it's just a http request maker.

PR #439, opened 2019-11-29
Add soft delete (discard) for User model

### LINKED ISSUE #385: Prevent user from deleting account if unpaid invoices issued
At the moment user can destroy their account without any obligations. Destroying user should be prevented if they have won an auction and (unpaid) invoice issued.

yulgolem:
Closes #385 by adding soft-delete (discard) via [discard gem](https://github.com/jhawthorn/discard) on User model.


app/models/user.rb
```diff
@@ -35,6 +37,18 @@ class User < ApplicationRecord
...
+    super && !discarded?
+  end
+
+  def can_be_discarded?
+    !cannot_be_discarded?
+  end
+
+  def cannot_be_discarded?
+    invoices.issued.present?
+  end
```
karlerikounapuu: What's the purpose of that? Why not combine into single method?

yulgolem: Just semantic sugar. I'll change it to single, 'cause the main condition is 'if it can be discarded'

artur-beljajev: Perhaps `discardable?` is more eloquent?

artur-beljajev: Is there really a need to use gem for such simple feature?

yulgolem: discardable? looks good for me.
As for gem - just trying not to reinvent a bike every time.

vohmar:
Artur just pointed out that we already have an audit schema set up for auctions meant to be used for archiving  deleted data. So the question is maybe its not necessary to introduce new  soft delete logic here as similar audit/arhivation model is used in our other apps as well - a bit differently sure (only auction has it set up with separate db schema)

yulgolem:
Well, since User audit via triggers and `audit` namespace is already implemented, the question is - do this PR is needed at all and maybe we should close the issue since deleted users data os already stored?

vohmar:
as user deletion is already handled by audit log solution with records remaining in separate audit db schema I am closing this PR and respective issue ticket. Sorry for the confusion and wasted time and effort.

vohmar:
the soft delete gem was only part of the issue - main point that is to prevent users with unpaid invoices to be deleted should still be solved.

yulgolem:
This PR and soft-delete functionality are not needed.

Issue #289, opened 2019-06-14
Admin: Add turn count to auction list


vohmar:
Add counter of the number of rounds the same domain has been cycling through auction process without being released or registered

statuses can be used to determine the fate of an auction
no bids - domain no bids were made and domain was released - cycle ended
domain_registered - domain was registered - cycle ended
payment_not_received - bid was made, but the invoice was not paid - new round
domain_not_registered - bid was made, invoice was paid, but domain was not registered - new round

other statuses indicate ongoing auction process

(Martin has script to calculate this)

This is needed so we could track and spot if some domains are circling around for too long indicating possible abuse.

vohmar:
matrin's query: select domain_name, count(domain_name) from auctions group by domain_name having count(domain_name) > 1 order by count, domain_name;

it must be modified a bit as this one does not take into account the possibility that the domain name was registered in between (that should reset the count on next auction).

yulgolem:
After deploy run `rake data_migrations:count_auction_turns `

vohmar:
fixed by #450

PR #450, opened 2019-12-12
Add count of auction rounds on domain name to auction

yulgolem:
See #289

### DIFF
--- a/app/models/auction.rb
+++ b/app/models/auction.rb
@@ -107,20 +107,12 @@ class Auction < ApplicationRecord
     end
   end
 
-  def in_progress_by_date?(date)
-    (starts_at <= date.end_of_day || starts_at <= date.beginning_of_day) &&
-      (ends_at >= date.end_of_day || ends_at >= date.beginning_of_day)
-  end
-
-  def search_data
-    {
-      id: id,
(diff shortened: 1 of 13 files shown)


yulgolem:
After deploy run `rake data_migrations:count_auction_turns `

app/models/auction.rb
```diff
@@ -107,20 +108,23 @@ def in_progress?
...
-      id: id,
-      domain_name: domain_name,
-      created_at: created_at,
-      starts_at: starts_at,
-      ends_at: ends_at,
-      offers_count: offers.count,
-      completed: result.present?,
-    }
+  def calculate_turns_count
+    auctions = Auction.unscoped.where(domain_name: domain_name).where('starts_at <= ?', starts_at)
```
karlerikounapuu: Wondering if this could be moved to scope instead so it can be used by potential other developments? Or implementing 2 params for scope function would be too "hacky"?

yulgolem: Not 'hacky' at all, but I prefer not to pre-improve.
Suppose we'll move it to scope when it will be any other place it needed. Until when make a scope for a single usage seems like overkill to me.

PR #453, opened 2019-12-16
382 consider settings refactoring

### LINKED ISSUE #382: Consider Settings refactoring
As Karl mentioned before, we're already have two issues (#372 and #356 ) that shows us that the current structure of the Setting object does not suit us well.
We cannot change type of setting stored if we need some kind of serialization, etc, etc.

So - why don't we create some kind of custom settings object not in Entity-attribute-value paradigm, but with classic relation DB approach, where one column will be equal one or more settings (maybe via store_accessor, stored in json instead of text... but it's unnecessary details atm), and one table entry will be equal the whole settings set for application?

We might leave Setting class as a settings wrapper for some kind of backward compatibility as well.

@karlerikounapuu , @artur-beljajev your thoughts?

karlerikounapuu:
Closes #382 

New behaviour of settings

- Every setting is now typed (string, integer, boolean, hash, array). These data types are stored in ApplicationSettingFormat model.
- Each ApplicationSettingFormat has a json hash of ApplicationSetting(formerly Setting) in `settings` field.
- Settings must be queried via `ApplicationSetting.setting_code` (formerly `Setting.setting_code`)

Setting can be edited via `ApplicationSettingFormat.update_setting!(code:, value:, description:)`

WIP


app/models/application_setting_format.rb
```diff
@@ -0,0 +1,63 @@
+class ApplicationSettingFormat < ApplicationRecord
+  store_accessor :settings
```
yulgolem: What's the point in using `store_accessor` here?

karlerikounapuu: A leftover 😀 Thanks for pointing it out

karlerikounapuu:
This PR is not relevant anymore since we came up with new solution.

PR #468, opened 2019-12-27
Add value format to Settings

### LINKED ISSUE #382: Consider Settings refactoring

karlerikounapuu:
Closes #382 

- Every setting is now required to have `value_format` which defaults to `string`. Available formats are `string`, `boolean`, `integer`, `hash`, ` array`

- Settings are no longer accessed via Setting.setting_name because it has too many redundant methods.

- Each invidual setting must be retrieved via `Setting.find_by(code: :setting_code).retrieve`. It has same output as previous querying via class method.

`data_migrations:set_correct_setting_formats` rake task must be ran to set correct format for already existing settings.

### DIFF
--- a/app/models/setting.rb
+++ b/app/models/setting.rb
@@ -3,37 +3,59 @@ class Setting < ApplicationRecord
   validates :code, presence: true, uniqueness: true
   validates :description, presence: true
   validates :value, presence: true
+  validates :value_format, presence: true
+  validate :valid_value_format
 
-  def self.auction_currency
-    Setting.find_by(code: :auction_currency).value
+  VALUE_FORMATS = {
+    string: :string_format,
+    integer: :integer_format,
+    boolean: :boolean_format,
+    hash: :hash_format,
+    array: :array_format,
+  }.with_indifferent_access.freeze
+
+  def retrieve
+    method = VALUE_FORMATS[value_format]
+    send(method)
   end
 
-  def self.auction_minimum_offer
-    Setting.find_by(code: :auction_minimum_offer).value.to_i
+  def valid_value_format
+    formats = VALUE_FORMATS.with_indifferent_access
+    errors.add(:value_format, :invalid) unless formats.keys.any? value_format
   end
 
-  def self.terms_and_conditions_link
-    value = Setting.find_by(code: :terms_and_conditions_link).value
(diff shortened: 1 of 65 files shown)


app/models/setting.rb
```diff
@@ -3,107 +3,66 @@ class Setting < ApplicationRecord
...
-    Setting.find_by(code: :registration_term).value.to_i
+  def boolean_format
+    value == 'true'
   end
 
-  def self.auction_duration
-    value = Setting.find_by(code: :auction_duration).value
+  def hash_format
+    return JSON.parse(value)[I18n.locale.to_s] if code == 'terms_and_conditions_link'
+    return JSON.parse(value)[I18n.locale.to_s] if code == 'violations_count_regulations_link'
```
yulgolem: How about to make a separate method for this?
Something like `localised_value`? We are returning the same value of hash in two different conditions - maybe we could join them somehow?

app/models/setting.rb
```diff
@@ -3,107 +3,66 @@ class Setting < ApplicationRecord
...
 
-    Setting.find_by(code: :auction_duration).value.to_i
+    JSON.parse(value)
+  end
+
+  def array_format
+    JSON.parse(value)
```
yulgolem: .to_a? Maybe we should enforce array as a return here?

karlerikounapuu: Doesn't really change current behaviour, but sure, for better clarity I'm willing to add it 🙂

app/models/setting.rb
karlerikounapuu: Might be even better. Thanks for suggestion.

PR #481, opened 2020-01-09
Add app.json for Heroku compatiblity, add database and setting configs

yulgolem:
Heroky preview app setup.

See #502


yulgolem:
#502 blocks from merge.

PR #509, opened 2020-02-07
Remove wkhtmltopdf-binary gem

karlerikounapuu:
Removed wkhtmltopdf-binary from Gemfile as it's included now in docker image.

Modified Travis conf to include wkhtmltopdf binary in test env


vohmar:
receiving true.html (download failed) when trying to download an invoice (regular user, my invoices, download). Same issue with admin. Does it require configuration?

PR #521, opened 2020-02-18
Directo: Force :en locale

### LINKED ISSUE #518: Directo: invoice sums are cent in cents
we are sending all invoice sums from auction in cents as is in db, but Directo expects EUR value.
Need to divide all the values by 100 before forwarding these to Directo.

karlerikounapuu:
Closes #518 

Most likely the reason why live env failed to assign correct invoice pricing was because of locale.
Turns out that directo allows only points in price (10.00), but live tried to send with commas (10,00)


teadur:
Included en_US.UTF-8 in ruby baseimage, https://github.com/internetee/docker-images/pull/7/

Issue #513, opened 2020-02-11
Auto-run data migrations on deploy


yulgolem:
We need some way to automatically run all previous/actual data migrations as per usual DB migrations.

Possible variants:
1) move data migrations to db migrations, on separate classes.
2) make a rake-task which automatically run all data migrations in order of timestamps (need to change name structure) and add this task to deploy script.

@vohmar , @karlerikounapuu Any ideas?

karlerikounapuu:
In order to run data migrations automatically, we should somehow implement some simple versioning for data migrations as well as current out-of-box solution for regular migrations. I believe our current data migrations are not reversible nor designed for rerunning multiple times. They're meant for "run once" behaviour.

I would go with solution no 2. We could also create post-hook for db:migrate task via `enhance` so it runs that custom rake task in addition to db:migrate without custom deploy script tuning.

yulgolem:
Let's consider this gem for versioning/migrating - https://github.com/ilyakatz/data-migrate

teadur:
I proposed we use data-migrate in registry aswell so going with same solution for both makes sense :)

https://github.com/internetee/registry/issues/1298

yulgolem:
Need to add `bundle exec rake data:migrate` to deploy scripts to run data migrations.
Already added it to travis.yml.

vohmar:
the #525 is merged to master and to close this issue the deploy scripts need to be updated. @teadur @ratM1n please do so.

vohmar:
fixed by #525

PR #525, opened 2020-02-21
Add data-migrate gem

yulgolem:
See #513


lib/tasks/data_migrations/convert_terms_and_conditions_link_setting.rake
```diff
@@ -2,16 +2,17 @@ namespace :data_migrations do
   task convert_terms_and_conditions_link_setting: :environment do
```
yulgolem: Suppose maybe we should get rid of all the old data migration, how do you think?

karlerikounapuu: That's what I was thinking also. They don't seem to be relevant anymore and I doubt someone has few months old registry running somewhere in the closet

PR #535, opened 2020-03-09
Directo: Comply with new Directo gem structure

karlerikounapuu:
Do not merge before Directo gem changes are not moved from `directo-api` to directo `master` branch.

Requires Gemfile changes in `auction` and `registry` project at the same time.


vohmar:
invoices reached directo as expected.

vohmar:
tested, good, ready to merge

Issue #561, opened 2020-06-05
Investigate websockets to enable English auction


maciej-szlosarczyk:
Investigate ActionCable usage and then estimate how much time it would take to develop functionality to inform live users on a website that they have been outbid.

Required for English auction.

maciej-szlosarczyk:
Things to figure out/configure:

* Websockets need the proxy server to support upgrade requests. It should not be a problem, but it will take 1-2 days to get it right.

maciej-szlosarczyk:
Another thing: Because Ruby/Rails is slightly backwards and needs threads, there probably needs to be Redis included in the production setup if we want to publish socket information from background jobs.

OlegPhenomenon:
@vohmar close this issue, please

OlegPhenomenon:
what's the point of reopen this tickets if websockets already implemented? it was disable until Martin will install redis on test production and production envs?

PR #562, opened 2020-06-05
Very basic "console.log" presentation of websockets

maciej-szlosarczyk:
### Please do not merge, this is only for demonstration

On every load of users/UUID page, a message is broadcasted to all
users with some arbitrary data inside (even if they browse a different page)

The good:

Can be used to broadcast messages about auctions, either by auction
UUID or domain name. Every user who bid to that auction and is
browsing the website, can get an UI alert that the auction has ended
or the autobidder functionality has been trigger and current price is
now higher. Broadcasting is fairly straighforward, see example in
users_controller#show.

Routing and making decision who to send a message can be completely
handled by Rails side, there is no need to write any complex
javascript logic or expose additional data as JSON.

The bad:

* Will need Redis for broadcasting messages from background jobs,
because Ruby and threads.
* Will need some production configuration on proxy level to handle
upgrade requests. Locally, I exposed the puma server directly and
created a socket bound to that port.

FYI @yulgolem @karlerikounapuu

### DIFF
--- a/app/channels/application_cable/connection.rb
+++ b/app/channels/application_cable/connection.rb
@@ -1,4 +1,17 @@
 module ApplicationCable
   class Connection < ActionCable::Connection::Base
+    identified_by :current_user
+
+    def connect
+      self.current_user = find_current_user
+      logger.add_tags 'ActionCable', current_user.email
+    end
+
+    def find_current_user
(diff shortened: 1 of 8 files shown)


maciej-szlosarczyk:
Relates to #561

OlegPhenomenon:
no actual

PR #574, opened 2020-08-04
Add searching by invoice number

### LINKED ISSUE #523: Search invoices by invoice nr
In admin it is currently not possible to search invoices by invoice nr. It is necessary for accountign auditing for example

User story: Accountant notices that invoice sum in accounting system does not match the sum of the payment made for this invoice. Accountant asks auction admin to verify if the data of an invoice nr XXXXXX is correct in the accounting system. Auction admin logs into auction admin interface goes to invoices view and enters the invoice nr as a search criteria.

yulgolem:
Closes #523

### DIFF
--- a/app/controllers/admin/invoices_controller.rb
+++ b/app/controllers/admin/invoices_controller.rb
@@ -1,3 +1,4 @@
+# rubocop:disable Metrics/ClassLength
 require 'invoice_already_paid'
 
 module Admin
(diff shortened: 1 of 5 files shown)


vohmar:
getting internal error if the search term begins with a number but contains any other characters except numbers ie 1a

yulgolem:
@vohmar Fixed, covered by test.

PR #580, opened 2020-09-16
Update rails & yarn packages

yulgolem:
Updates rails gem & yarn packages to address security issues.


yulgolem:
Have some errors compiling assets on staging, locally on container all works well. @ratM1n please, check staging deploy, maybe we'll need to rebuild node-sass(`npm rebuild node-sass`)

ratM1n:
@yulgolem Hi! deployed this branch to test, deploy went seamlessly but portal has assets problem ... i get errors from language selection and logout -buttons ... probably login with TARA doesn't work also. Seems same error as we had with prod.

PR #631, opened 2021-04-08
Banned users bids should be cancelled

### LINKED ISSUE #610: Banned users bids should be cancelled
If user gets banned from participating in any auctions but has managed to get in a bid to an active auction (multi day auction for example) the user ends up in a situation where they cannot change nor cancel the existing bids.

The correct action would be to automatically cancel all the active bids for the user as soon as the complete ban is activated. The same goes for the single auction ban - if a ban is activated the active bids for the same domain name should be cancelled.

OlegPhenomenon:
Close #610

### DIFF
--- a/app/models/ban.rb
+++ b/app/models/ban.rb
@@ -5,9 +5,6 @@ class Ban < ApplicationRecord
   validates :valid_until, presence: true
   validate :valid_until_later_valid_from
 
-  # after_create :remove_offer_from_active_auction
-  # after_create :remove_offers_from_active_auctions
-
   def valid_until_later_valid_from
     return unless valid_until
     return if valid_until > (valid_from || Time.zone.now)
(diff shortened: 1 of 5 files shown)


app/models/ban.rb
```diff
@@ -19,4 +22,26 @@ def valid_until_later_valid_from
...
+    return unless auction.in_progress?
+
+    user = User.find(user_id)
+    offer = user.offers.find_by(auction_id: auction.id)
+    offer.destroy if offer.present?
+  end
+
+  def remove_offers_from_active_auctions
+    ban_number_of_strikes = Setting.find_by(code: 'ban_number_of_strikes')
+    user = User.find(user_id)
```
yulgolem, line 38: Check `user.completely_banned?` method, it was intended to do like so.

app/models/ban.rb
```diff
@@ -5,6 +5,9 @@ class Ban < ApplicationRecord
   validates :valid_until, presence: true
   validate :valid_until_later_valid_from
 
+  after_create :remove_offer_from_active_auction
```
yulgolem, line 8: Also I suggest to move all the logic from AR callbacks on model to AutomaticBan, where all the bans are created.

yulgolem (changes requested):

PR #854, opened 2021-11-23
Everypay upgrade

### LINKED ISSUE #848: Upgrade to Everypay APIv4
in november 2021 the support for v3 of the everypay API will end so all integrations should be upgraded to v4.

doc:
https://support.every-pay.com/downloads/everypay_apiv4_integration_documentation
https://support.every-pay.com/ms/integration/

v4 enables using linkpay and open banking features so we can drop using swedbank's and seb's banklinks and direct all payments to our primary bank account. WE can also add linkpay feature to all emails and pdf invoices.

DinSmol:
close #848


vohmar:
cannot get it properly working still - different issues from unhandled callbacks to errors like Uncaught TypeError: Cannot read properties of undefined (reading 'forEach'). Will close this issue once the new billing service is in production.

vohmar:
new billing service in prod, works on api v4

PR #870, opened 2022-01-24
Integrate billing system

OlegPhenomenon:
Reminder:
I made two migrations, one migration creates a new `number_two` column in the `invoice` table, after which the migration copies all number values into `number_two`

The second migration removes the default number auto-increment from the invoice table to allow the billing system to assign the correct number to the invoice itself.

I took these steps for security reasons, so as not to lose data in case of an unforeseen situation.

When everything is already working as it should, then the number_two column in the invoice table can be deleted

Migrations file names:
20220127103133_add_number_two_column_to_invoice.rb
20220127105212_change_default_number_value_to_invoice.rb


OlegPhenomenon:
Useless code

PR #878, opened 2022-02-15
integrate billing system

OlegPhenomenon:


vohmar:
* Payment button - please remove everypay logo and replace it with just pay / maksa lables - most people do no know what everypay means or is about. They just want to pay - lest make it clear for them
* the two step process of paying multiple invoices should be streamlined - lets put the invoice generation and payment into one. So users presses the payment button in the background invoice is generated and user is directed to payment inteface. At the moment its confusing as user does not see the newly generated invoice and probably will not understand what this is for anyway. Additionally when user leaves the view or refreshes it the payment button is gone and user has to genrate new invoice to make the payment.
image

PR #884, opened 2022-02-28
Fixed ordering and sorting issues of admin lists

### LINKED ISSUE #883: Fix admin lists issues with sorting and ordering data
After fixing issue #873 with sorting out invoices, found out that all the models in admin section have the same sorting and ordering problem, cancelling each other. Search and index pages have WET code

maricavor:
Close #883 

- Fixed sorting & ordering issues of admin lists
- Refactored code of search and index pages, made DRY
- Eliminated n+1 issues
- Added missing healthcheck translations


OlegPhenomenon (approved):

app/controllers/admin/results_controller.rb
```diff
@@ -24,15 +24,17 @@ def create
 
     # GET /admin/results/search
     def search
-      search_string = search_params[:domain_name]
-      statuses_contains = params[:statuses_contains]
+      @_params = search_params.compact_blank
```
OlegPhenomenon, line 27: If you do not plan to use class variables in the view, then it is better not to do so. Because, I expect (or someone else) that they are also used in view and logic is built on them, but in reality they are not there.

app/controllers/admin/results_controller.rb
```diff
@@ -78,15 +80,16 @@ def search_by_users_ids(user_ids)
 
     def search_domain_names_result
       Result.joins(:auction)
-            .includes(:offer, :invoice)
-            .where('auctions.domain_name ILIKE ?', "%#{@origin}%")
+            .includes(:auction, :offer, :invoice)
```
OlegPhenomenon, line 83: Why include the auction here, if it is already loaded at the SQL level through the joins method? It has already been merged with the results table, so there is no point in using it for preloading via include to avoid N + 1.

OlegPhenomenon (changes requested):

app/controllers/admin/results_controller.rb
maricavor: :joins joins tables together in sql, :includes eager loads associations to avoid the n+1 problem (where one query is executed to retrieve the record and then one per association which is loaded). It will not work here if I remove one of these functions.

OlegPhenomenon (approved):

OlegPhenomenon:
deprecated

PR #900, opened 2022-04-27
English Auction

OlegPhenomenon:
Need to run this when deployin to assign auction type to all auction in registry auctions db that are lacking it: lib/tasks/assing_auction_platform_type.rake


vohmar:
Deposit functionality:
* admin: why two separate chekcboxes enable and disable deposit? how about just one enable deposit - by default unchecked and admin can check it or uncheck it by just clicking
* If possible replace the bid button in the auction and auction list view with osale / participate button in case the auction requires deposit and user has not made it yet (is not enabled to participate in the auction)
* Edit the deposit explanation: "To participate in the auction, you need to make a deposit of 500.00 (VAT included) euros. If you win the deposit will be your prepayment. Deposit is returned to other participants within a week after the end of the auction."
* Do not display all users in the participant list of the English auction - instead display only those that have clicked on the make a deposit button and perhaps add a search box to look for users from the general users list so admin could add the user and enable the participation if something went wrong with the automated process

PR #902, opened 2022-05-06
Add wishlist automatic bids

### LINKED ISSUE #893: Whishlist: add optional price field
Add an option to add a price to a wishlist entry, so when a domain is put on auction the price is offered automatically for the user unless access is restricted (ban for violation(s)).

in case of blind auctions the offer is made asap
in case of english auctions (incl secondary market) this is the upper limit setting for the autobidder

thiagoyoussef:
close #893


OlegPhenomenon (approved):

vohmar:
* please add enter button at the end of each wishlist row to provide users addition clicking/tapping option in addition to current "enter to confirm"
* actions column is not translated into Estonian - use "Toimingud"
* the wishlist prices are not shown in Estonian view

OlegPhenomenon (approved):

PR #930, opened 2022-09-12
remove banklinks

### LINKED ISSUE #924: Remove banklinks
Everypay provides oneoff payments that can return payment states. The problem with using banklinks was that for the test environment, callbacks came with a long delay, and clients who used the test environment were not very happy. Oneoff endpoint campaign solves this problem. If everything works as expected on the test environment, then banklinks become unnecessary and they will only take up code.

OlegPhenomenon:
close #924


vohmar:
getting strange error if trying to pay overdue invoice
image

vohmar:
this is fine by me after resolving merge conflicts and if possible then then replacing "prepending" description in the payment info with something more meaningful like the domain name for example

image

OlegPhenomenon:
deprecated

PR #931, opened 2022-09-14
fix eid sign up

### LINKED ISSUE #927: Adding new eid user results in missing password error
When loggin in to auction protal using eid first time then user is met with error message about blank password field even though there is no such option provided in that view

Ther error message about blank password cannot be there as this is in no way mandatory in case of eID authenticated users. Lets no add the password field to that view to keep the onboarding simple and clean.

image

thiagoyoussef:
close #927

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -169,7 +169,7 @@ class User < ApplicationRecord
     user.given_names = omniauth_hash.dig('info', 'first_name')
     user.surname = omniauth_hash.dig('info', 'last_name')
     user.country_code = uid.slice(0..1)
-    user.save
+    user.save if user.valid?
 
     user
   end


app/models/user.rb
```diff
@@ -169,7 +169,7 @@ def self.from_omniauth(omniauth_hash)
     user.given_names = omniauth_hash.dig('info', 'first_name')
     user.surname = omniauth_hash.dig('info', 'last_name')
     user.country_code = uid.slice(0..1)
-    user.save
+    user.save if user.valid?
```
OlegPhenomenon, line 172: but what to do if user invalid?

OlegPhenomenon (changes requested):

PR #937, opened 2022-09-26
Allow to delete auctions when does not have any offers

### LINKED ISSUE #932: Deleting auction in admin does not work
image

thiagoyoussef:
close #932

### DIFF
--- a/app/models/auction.rb
+++ b/app/models/auction.rb
@@ -86,7 +86,7 @@ class Auction < ApplicationRecord
 
   def can_be_deleted?
     if valid?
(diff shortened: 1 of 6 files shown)


vohmar:
works as expected - well done!
Please update the error message "You cannot delete an auction that is already in progress or has finished.":
en: "It is not allowed to delete an action that is already in progress and has bids or has alrady finished"
et: "Oksjonit, mis on alanud ja millele on tehtud pakkumine või mis on juba lõppenud, ei saa kustutada."

maricavor (approved):

PR #949, opened 2022-10-25
946 incorrect vat calculation

### LINKED ISSUE #946: Incorrect vat calculation
Rules:
1. Any user/billing profile marked in Estonia  = 20% vat
2. any user/billing profile marked outside of EU = 0% vat
3. Any VAT liable (vat nr present in the billing profile) EU org = 0% vat (actually this is reverse VAT, but for the simplicity we use 0)
4. Any non VAT liable org and all priv billing profiles from EU = vat rate of that EU country 

issues:
* private billing profile from GB - 20% vat applied to the invoice in auction/billing, correct would be 0 as non-EU (rule 2); invoice sent to directo was correct with 0% vat

Other scenarios seemed to works as expected. GB issue might be caused by the brexit as it used to be part of EU, but not any more

OlegPhenomenon:
close #946

### DIFF
--- a/app/controllers/billing_profiles_controller.rb
+++ b/app/controllers/billing_profiles_controller.rb
@@ -81,9 +81,6 @@ class BillingProfilesController < ApplicationController
     update_params = params.require(:billing_profile).permit(
       :name, :vat_code, :legal_entity, :street, :city, :postal_code, :country_code
     )
-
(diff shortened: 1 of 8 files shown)


vohmar:
vat calculations for GB are now correct, but the invoice arrived to Directo with currency MXN that is incorrect

image

image

PR #963, opened 2022-11-15
admin order fix

OlegPhenomenon:


vohmar:
for some reason NaN is listed as highest possible value - so if i order auctions by bid from highest to smaller i get first the auctions with no bids (NaN). Its should be the other way around - NaN is the smallest possible value

PR #979, opened 2022-12-19
domain history

OlegPhenomenon:


vohmar:
Nice, but usability is tricky. Could we put current and finished auctions all under auctions menu different tabs/sub-menu options or add filtering options there. With filtering option showing and searching by default only current auctions, but if user enables can return all or only finished auctions.

vohmar:
Few more simple change requests:
* in en view please change the first letter into capital for "list all auctions" and "list 15 auctions per page"
image

* in et view please rename the "Lõpetatud oksjonid" to "Lõppenud oksjonid"

* How about moving the "Praegused oksjonid" / "Current auctions" link to the right side of the table - same place where in the current view is the "Finished auctions" link?
image

PR #981, opened 2022-12-20
fixed admin ordering for finished auctions

OlegPhenomenon:


vohmar:
but why no sorting by price and winner?

PR #996, opened 2023-01-17
fixed invoice results with deposit calculation

### LINKED ISSUE #990: Deposit not considered as prepayment
Branch: fixed-auction-creator

deposit is a form of prepayment - if deposit is required and paid it needs to be taken into account as a prepayment for the winner of the auction meaning that it should be shown on the auction invoice and the payment required needs to be less by the deposited amount ie 

deposit: 500€
winning bid: 600€
invoice. 600€, deposit -500€, due to be paid 100€

for overdue invoices it should be the same - as we do not return deposit to auction winners paying the invoice after the fact to lift the ban must also consider the deposited amount.

OlegPhenomenon:
close #990

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+31.01.2023
+* added list of finished auctions https://github.com/internetee/auction_center/issues/978
+
 20.01.2023
 * fixed ban count issue https://github.com/internetee/auction_center/issues/988
 
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -30,6 +30,9 @@ class Invoice < ApplicationRecord
 
   before_create :set_invoice_number
 
+  delegate :enable_deposit?, to: :enable_deposit?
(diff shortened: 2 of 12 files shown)


vohmar:
lets change the invoice view a bit:
1. let's leave only the bid listed as an invoice item - the reason being that we need to calculate the vat amount from this sum and only then we can deduct deposit as prepayment
2. lets add the deposit as add two new lines after the total line - deposit and amount due (et: ettemaks, tasuda)

vohmar:
invoice calculation is off - it adds deposit sum to the invoice total :
image

translation issue in English version:
image

could we make it look like something like this:
image

the deposit and amount due needs to get into pdf and e-invoices as well
image

vohmar:
looks good, but admin has still old invoice template:
image

vohmar:
calculation error on the invoice in the admin portal
image

there is no such issue with the same invoice in the user's view. Downloaded pdfs are fine in both cases.

PR #997, opened 2023-01-17
deposits should be cancelled when auction comes to the next round

### LINKED ISSUE #991: Cancel deposit when auction finished
When auction with enabled deposit requirement ends the deposit has to be released for all the participants that did not win the auction

it's not the case at the moment and if the auction comes back to n-th round all the participants enabled to participate in previous round are automatically enabled for new round including the banned winner (ban prohibits to make the bid, but user can access the bidding view as if the deposit is paid)

OlegPhenomenon:
fixes #991

### DIFF
--- a/app/models/registry/auction_creator.rb
+++ b/app/models/registry/auction_creator.rb
@@ -97,6 +97,8 @@ module Registry
         auction.starting_price = legacy_auction.starting_price
         auction.min_bids_step = legacy_auction.starting_price
         auction.slipping_end = legacy_auction.slipping_end
+        auction.enable_deposit = legacy_auction.enable_deposit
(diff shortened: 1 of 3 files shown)


vohmar:
The deposit from last round is not taken into account any more, but the users that participated in the previous round are not able to make new deposit payment - the payment goes through, money is taken but the participate button remains in place not allowing to make any bids but accepting endless deposit payments

steps to reproduce:
1. create english auction
2. make the deposits and bids
3. finish the auction
4. make the payment 
5. set the registration deadline to the past
6. update results - status should change to not registered
7. run auction creation job to restart the auction
8. make the deposit payments with the users from the previous round
9. see if you can make a bid after successful deposit payment

PR #998, opened 2023-01-18
added deposit management to admin

### LINKED ISSUE #992: Deposit management
* add paid deposit list in admin - list all received deposit payments (user, date, sum, auction, status)
* add deposited payment info to user view (under my invoices?) (date, sum, auction, status, refund time)
* in billing it should be possible to filter out deposit payments - perhaps add new status "deposit" or "deposited" as there are no invoices created for those payments
* in Everypay its also necessary to differentiate deposit payments from others - add special order nr range for auction deposit payments
* automatic deposit refund after the end of the auction? - possible via Everypay API. Doc: https://support.every-pay.com/merchant-support/api/
* replace the "tehke deposiit" button name in the deposit view with "Tasuge deposiit"

OlegPhenomenon:
close #992

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,6 +33,7 @@ gem 'skylight'
 gem 'sprockets', '~> 4.0'
 gem 'turbolinks', '~> 5'
 gem 'webpacker', '~> 6.0.0.rc.5'
+gem 'i18n-debug', group: :development
 
 # token
 gem 'jwt'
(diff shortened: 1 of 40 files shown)


vohmar:
* i was able to make only on successful deposit payment. After that all will end with error "{"code"=>4024, "message"=>"The order reference '10003' has already been successfully paid"}". The invoice record is created in billing but its not paid. On every attempt new record is created in billing, and the user receives the same error on traying to pay

* rename the deposits section in "your invoices" view to "Paid deposits"
* estonian translation for Paid deposits - "Deposiidi maksed"
* the deposit records should be ordered by date from newst to oldest (in both user and admin views)

PR #1015, opened 2023-02-17
Added ability for banned users for making deposits

### LINKED ISSUE #1009: Banned users can pay the deposit
Auction system lets users that are banned from participating in the auction to pay the deposit and after that deny making the bid.
Users banned from participating in any or all of the auctions should also be banned from making the deposit payment for the auctions they are not authorised to participate in.

maricavor:
Close #1009 

- Updated .gitignore
- Small code refactoring

### DIFF
--- a/app/controllers/invoices_controller.rb
+++ b/app/controllers/invoices_controller.rb
@@ -32,7 +32,8 @@ class InvoicesController < ApplicationController
 
     return unless params[:state] == 'payment'
 
-    message = 'Payment was successfully created. Your payment will be processed as soon as possible. Thank you!'
+    message = 'Payment was successfully created. ' \
+      'Your payment will be processed as soon as possible. Thank you!'
     flash[:notice] = message
   end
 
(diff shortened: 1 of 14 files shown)


app/controllers/invoices_controller.rb
```diff
@@ -32,7 +32,8 @@ def index
 
     return unless params[:state] == 'payment'
 
-    message = 'Payment was successfully created. Your payment will be processed as soon as possible. Thank you!'
+    message = 'Payment was successfully created. ' \
+      'Your payment will be processed as soon as possible. Thank you!'
```
OlegPhenomenon, line 36: I forgot to correct one flaw here. It would be better if this message was localised.

OlegPhenomenon (changes requested):

app/controllers/invoices_controller.rb
maricavor: Done!

PR #1016, opened 2023-02-20
upper limit for bids && tests

OlegPhenomenon:


vohmar:
* tired to make bid of 21500000€ for auction aba6.ee at auction staging - internal error

bid step after 100000€ is always 10000€ - that is OK.

vohmar:
Please change the error message on exceeding the max bid limit:

en: "Your bid is denied - maximum bid size exceeded!"
et: "Pakkumine ebaõnnestus! Pakkumine ületab süsteemi poolt lubatud maksimaalset summat!"

vohmar:
The prefix is still there:
image

image

PR #1023, opened 2023-02-28
added offer and auction notifications

OlegPhenomenon:


vohmar:
Some copywriting fixes to the web push confirmation popup:
en: "Web Push: The .ee auction portal employs web push notifications to keep you informed of any over-bidding activity, without requiring you to keep the website open constantly."
et: "Veebi teavitused: .ee oksjonite portaal kasutab veebi teavitusi, et hoida teid kursis, kui teie pakkumine on üle pakkutud, ilma et peaksite kogu aeg veebilehte avatuna hoidma."

The agree button on the popup message does not do anything wile the disagree or the close button at least close the window.

Similar issue is on the profile view - the "webpush available" button does not seem to do anything

no notification is shown if user is in some other tab  as the web push description suggests

the bell icon: messages about new bids appear there even if user has autobidder set and user is still the highest bidder. Create the message only if user is not the highest bidder any more keeping the autobidder option in mind. And only create a message when the state changes for that user. Now a new message is created for all users that have made at least one bid for the auction every time someone makes a bid creating a lot of "spam"

also tests for ubuntu-20.04, 3.0.2 and 2.7 are failing.

vohmar:
That web push thing is mighty strange creation - for instance it gave me a notification even when the user session was timed out. The confirmation behaves strangely - sometimes i did get the browser confirmation when enabling the web push, but mostly I did not get it, but the push messages appeared regardless. For example in Edge the web push button in user profile was even inactive, but still the notification appeared from Edge as well.

would it be possible to make the notifications appear only if i was the highest bidder previously and now got overbid. Currently all participants get notified of any new bids to the auction

PR #1038, opened 2023-03-15
changed logic for ends_at and initial_ends_at

### LINKED ISSUE #1030: Getting new auctions from the registry is failing on staging
Attempts to get new auctions from registry, but it seems also updating results of finished auctions is failing on staging. Error in logs:

https://gist.github.com/ratM1n/64d24de65a210b42db81bfa5b2e9b5ce

auctions affected:
* billing-auction6.ee
* billing-auction7.ee
* bimba20.ee
* deposiit1.ee

OlegPhenomenon:
Close #1030 

I was unable to reproduce the error on the staging, but in this branch I have changed the logic of the `ends_at` assignment a bit. The point is that such an error could theoretically occur to auctions, which `initial_ends_at` field is not initialized, and auctions missed rounds for some reason or there was some delay. Therefore, before merging this branch, we need to test the entire logic of adding new domains to the auction and also test the addition of domains in the next rounds!

### DIFF
--- a/app/models/registry/auction_creator.rb
+++ b/app/models/registry/auction_creator.rb
@@ -85,14 +85,23 @@ module Registry
         auction.ends_at = nil
       else
         legacy_auction = auctions.first
-        new_ends_at = legacy_auction.ends_at
-        if legacy_auction.initial_ends_at.present?
-          legacy_time_difference = (legacy_auction.initial_ends_at - legacy_auction.starts_at).to_i.abs
-          legacy_difference_in_day = legacy_time_difference / 86_400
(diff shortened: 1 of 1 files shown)


vohmar:
sync from auction to registry does not seem to work - there are no errors in auction or registry logs, but there are errors in container logs:

Exception: ActionController::RoutingError: No route matches [GET] "/admin/jobs"

https://gist.github.com/ratM1n/e6b8d6814dcfcec65c2a80ecd595c917

the errors seem to be generated each time:

* DomainRegistrationCheckJob
* ResultStatusUpdateJob

jobs are run

PR #1043, opened 2023-04-19
invoices with zero sum

### LINKED ISSUE #1042: Eng auc: handle situations where winning bid equals deposited sum
If the winning bid equals deposited sum no invoice is generated and the result remains in awaiting payment status.

If bid equals deposit the invoice for the bid sum must be created, the invoice sum is then reduced by the deposited sum and the amount due of the invoices will be 0 ie the invoice is automatically marked as paid and the registration code should be provided to the winner.

OlegPhenomenon:
close #1042 
with https://github.com/internetee/eis_billing_system/pull/74


vohmar:
in case winning bid matches the deposit
* invoice is now marked as paid in both auction and billing
! auction status is still "awaiting payment" though meaning that user cannot get access to the registration code even though the invoice is in status paid

PR #1075, opened 2023-05-10
added offer + tax information to the bid view

### LINKED ISSUE #1072: Add the offer+tax information to the bid view
in the new design we solved it like this:
image

so for the current design we might to something similar.
Instead of "The offer does not include VAT." we could print something like

en: Total x (+ y% VAT)
et: Kokku x (+ y% käibemaks)

PS: english version of the new offer view has the minimum bid already printed in the offer textbox, but it is missing from Estonian view
image

image

please add the minimum sum to Estonian view as well

OlegPhenomenon:
close #1072


vohmar:
i put the updated in the issue ticket

PR #1084, opened 2023-05-29
randomized auction index list

### LINKED ISSUE #1080: Make the default order of auctions in the index list view random
Current order of auctions in the auctions list is descending alphabetic. As a result all the English auctions that are currently running for numbers are listed at the very end of the list. TO give those auctions a fair chance lets order the list in random order and users can then filter or reorder it as they want or need.

OlegPhenomenon:
close #1080

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,5 +51,4 @@ yarn-debug.log*
 /yarn-error.log
 yarn-debug.log*
 .yarn-integrity
-
 /public/assets/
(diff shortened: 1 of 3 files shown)


vohmar:
ordering of the auction list does not work. 

* in public index page, not logged in: If i try to order the list by any column ie "domain name" it just refreshes the list and continues to show the list in random order (same with other columns)
* same for logged in user
* admin (/admin/auctions) has it a bit different - the list is not randomized and that is OK, but the ordering does not function properly here as well. Might not be a new problem

PR #1088, opened 2023-06-07
immutable invoice billing data

OlegPhenomenon:


vohmar:
pdf invoice includes alway the latest version of the assigned or default billing prodile if linked one is missing. This creates a situation where data on the screen is different than data on the pdf version. So if user changes something in the billing profile the paid invoices in the auction portal will have the previous billing profile version data, but downloaded pdf will have the updated info. Same happens if the billing profile is deleted, in that case pdf has the default billing profile data in it

once invoice is paid it is locked and the data should be immutable on screen and in the pdf

PR #1124, opened 2023-09-13
sort fixing

OlegPhenomenon:
**What do we have here?**
We have a sorting issue here. The problem is that things like the user's bid and username are being sorted incorrectly (actually, correctly), and there are also several issues with sorting in the admin panel. Here are the details: #1125 

**Why?**
The user's bids are actually sorted correctly, we just don't "see" it. The thing is, sorting is done by the Offer model, specifically by the 'cents' column. However, according to business logic, we only see bids made for the English auction, but we don't see bids made for the blind auction, even if they exist. So, we end up sorting the bids we see and the ones we don't, and that's why it seems to us that the sorting is wrong.

As for the fact that the user sorting doesn't occur - the answer is simple, it's just not implemented and another column of the table was specified instead.

Admin panel issue: domain names weren't sorted because of a simple typo in the column name, and sorting by deposit wasn't implemented at all.

The next issue was with class naming. The thing is, our records store record types in uppercase letters. For example, `PaymentOrder::LHV`, `PaymentOrder::SEB`. And these records expect there to be subclasses with the same names, specifically in uppercase. However, in Ruby on Rails, the naming logic is slightly different - it uses a capital letter followed by lowercase letters. So, the correct naming would be `PaymentOrder::Lhv`, `PaymentOrder::Seb`. Hence, there's a conflict with the naming. In the past, this all worked, but apparently, with the Ruby on Rails version update, this logic broke.

**Well, alright, what did you do?**
The main problem with implementing sorting was as follows: I can make it so that only English auctions are sorted, however, those users who placed a bid for blind auctions, they can see their bids. It turns out that in this implementation, only bids for the English auction will be sorted, and their bids for the blind will not be. Therefore, I had to complicate the SQL query so that bids for a specific user's blind auction were also included.

Implementing different business logic for sorting other columns also required SQL queries, but they are not as complex and convoluted as sorting by user bids.

I added exceptions to the inflections.rb file for the classes `PaymentOrder::LHV`, `PaymentOrder::SEB`, so Ruby wouldn't apply naming rules to them.

**How do we test all this?**
Testing is simple: just click on columns and ensure they are sorted correctly in alphabetical order. However, there are nuances for auction domain names for regular participants:

- [ ]     if you're a guest, you ONLY sort English auctions
- [ ]     if you're a participant but didn't place bids for the blind auction, you ONLY sort English auctions
- [ ]     if you're a participant, didn't place bids for the blind auction, but another user did, you ONLY sort English auctions
- [ ]     if you're a participant and placed a bid on a blind auction, you ONLY sort English auctions along with that blind auction where you placed a bid
- [ ]     if you're a participant and placed a bid on a blind auction, and another user also placed a bid on another blind auction, you ONLY sort English auctions and ONLY your blind auction.
- [ ]  It's necessary to ensure not only in the administrative part that the sorting works correctly but also that it works correctly for the user. 
- [ ] It's also necessary to check that the naming for `PaymentOrder` records for subclasses like `SEB` and `LHV` are in uppercase letters, and no others

### DIFF
--- a/app/controllers/admin/invoices_controller.rb
+++ b/app/controllers/admin/invoices_controller.rb
@@ -15,21 +15,9 @@ module Admin
 
     # GET /admin/invoices
     def index
-      sort_column = params[:sort].presence_in(%w[paid_through
-                                                 paid_amount
-                                                 vat_rate
-                                                 cents
-                                                 notes
-                                                 status
-                                                 number
-                                                 due_date
-                                                 billing_profile_id]) || 'id'
-      sort_direction = params[:direction].presence_in(%w[asc desc]) || 'desc'
-
       invoices = Invoice.accessible_by(current_ability)
                         .includes(:paid_with_payment_order)
                         .search(params)
-                        .order("#{sort_column} #{sort_direction}")
 
       @pagy, @invoices = pagy(invoices, items: params[:per_page] ||= 15)
     end
--- a/app/controllers/admin/invoices_controller.rb
+++ b/app/controllers/admin/invoices_controller.rb
@@ -1,4 +1,4 @@
-# rubocop:disable Metrics/ClassLength
+# rubocop:disable Metrics
 # require 'invoice_already_paid'
 
 module Admin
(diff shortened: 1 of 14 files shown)


vohmar:
admin-invoices:
* billing profile sorts according to profile id not profile name
* number sorting has null values listed as bigger than the ones with values - should be the other way around or alternatively leave these always to the end
* total:
image
image

* total paid - like with inv numbers the null values are sorted as highest instead of smallest
*

vohmar:
finished auction view - sorting does not work with any columns at all

maricavor (approved):

PR #1128, opened 2023-09-18
added refund support for card payments

### LINKED ISSUE #1121: Deposit and payment refund management
New processing accounts have been created in Everypay for deposit payments. These accounts enable reserving funds on auction participants bank account without actually transferring any. The pre-authentication delay is set for 7 days in test system meaning that registry has time up to 7 days to cancel the reservation before the payment is actually processed.

Relevant api endpoints: /charge, /capture, /void and /refund

/refund command works also for Open Banking payments to pay back deposit for auction participants that did not use card payment option

The processing account are
EUR3D3 for test
EUR3D4 for staging

OlegPhenomenon:
close #1121

### DIFF
--- a/app/jobs/capture_job.rb
+++ b/app/jobs/capture_job.rb
@@ -1,17 +1,17 @@
-class RefundJob < ApplicationJob
+class CaptureJob < ApplicationJob
   BASE_URL = AuctionCenter::Application.config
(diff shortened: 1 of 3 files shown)


OlegPhenomenon:
deprecated

PR #1140, opened 2023-09-21
added missing billing profile vat code to invoice

### LINKED ISSUE #1127: Invoice is lacking VAT code of the billing profile
The issuer data only includes Name and Address, but we should all VAT code as well if present

OlegPhenomenon:
closed #1127 

What happened?
Judging by this error #1127, the specific invoice view is missing the user profile's VAT code for which the invoice was generated.

What did you do?
I added a condition: if the user has a VAT code, it should be displayed in the view; if not, it should be omitted.

How to test?
Simply check the invoice: if the invoice is generated for a billing profile that has a VAT code, then this code should be displayed in the view. If not, it should be absent. Another test scenario could be to generate an invoice for a billing profile that has a VAT code, and then delete that billing profile. In either case, the VAT code should be displayed.

### DIFF
--- a/app/views/common/pdf.html.erb
+++ b/app/views/common/pdf.html.erb
@@ -215,21 +215,22 @@
                     <hr/>
                     <dl class="dl-horizontal">
                         <dt>
-                            <%= t('billing_profiles.name') %>
+                          <%= t('billing_profiles.name') %>
                         </dt>
                         <dd>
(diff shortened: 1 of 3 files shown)


vohmar:
the vat code is listed under issuer data, should be with the customer data

image

PDF is OK

PR #1151, opened 2023-09-26
Added e-invoicing feature to invoices

### LINKED ISSUE #1138: Add e-invoicing option
Auction portal lacks e-invoicing functionality. We have now had two instances in very short period of time with e-invoice requests for auction winners.

Add two buttons to the invoice view
* download e-invoice
* send e-invoice

Download option would just download the xml format invoice similarly to the PDF so user can import it to their accounting systems for instance

Send e-invoice option would forward the e-invoice to e-invoice gateway - like we do in registrar portal but only on the press of the button

maricavor:
- Close #1138


OlegPhenomenon (approved):

vohmar:
getting internal error when trying to open any invoice old and new in staging

Issue #1222, opened 2024-02-14
Incorrect vat rate set to invoice after payment


vohmar:
Auction user paid off overdue invoices to remove the bans from the user account.

Invoices were due in 2023
The payment was for the sum of 6€
Payment was accepted and invoice marked as paid

But invoice show now the sum of 6,1€ with 22% vat rate

All invoices issued prior to 2024 for Estonian customers should be with vat rate of 20%
* Need to fix the issue on settling the invoices
* Need to fix the vat rate on the invoices with 22% issued before 1st of Jan 2024
* Need to resend the fixed invoices to Directo

vohmar:
fixed by https://github.com/internetee/auction_center/issues/1395

PR #1223, opened 2024-02-15
solution for wrong vat rate for old invoices

OlegPhenomenon:
**What does this PR do?**
This PR relates to this task #1222 . The essence is that old invoices that were created before the year 2024 and have the country code Estonia in the billing profile change to the wrong VAT rate after payment in the current year. It turns out that the user pays at a VAT rate of 20 percent, but after payment, the invoice is marked as though it was paid at 22 percent, which creates confusion for accounting.
However, there is another problem, if, say, a user has an old unpaid invoice from the year 2023 and they decide to change the billing profile for this invoice, a VAT rate of 22 percent will be set if the billing profile has the country code of Estonia - this is also incorrect. This PR solves these two problems.

**But what about invoices that already have incorrect data?**
For this, you need to run the task `rake invoices:assign_values_to_old_invoices`. This task takes all invoices that were created before the year 2024, which have the country code Estonia and which have a VAT rate of 22 percent, it changes the VAT rate to 20 percent, and also sets `in_directo = false`. The directo task will have to be run manually.

**How to test?**

- First, you need to try to reproduce the errors, how this can be done:
- Finish an auction and generate an invoice - then you need to change the values (issue_date, due_date, created_at, vat_rate) to last year's in this invoice and try to pay it, after that check what the values of the invoices are now.
- You also need to try to simply pay a regular invoice to make sure that the changes didn’t spoil anything.
- You also need to generate an invoice with last year's values for (issue_date, due_date, created_at, vat_rate) and for this invoice change the billing profile, for example, from foreign to Estonian, and make sure that for the Estonian billing profile VAT rate = 20%.
- You also need to generate an invoice with last year's values for (issue_date, due_date, created_at, vat_rate), however, set the VAT rate value to 22 percent and change in_directo to true, then run the task `rake invoices:assign_values_to_old_invoices` and check the result.
- After running the above task, make sure that this task did not affect other invoices that were created in the current year.

### DIFF
--- a/app/controllers/payment_orders_controller.rb
+++ b/app/controllers/payment_orders_controller.rb
@@ -25,38 +25,6 @@ class PaymentOrdersController < ApplicationController
     @payment_order.callback_url = callback_payment_order_url(@payment_order.uuid)
   end
 
-  # # ANY /payment_orders/aa450f1a-45e2-4f22-b2c3-f5f46b5f906b/return
-  # def return
-  #   @payment_order = PaymentOrder.find_by!(uuid: params[:uuid])
-
-  #   if @payment_order.paid?
-  #     respond_to do |format|
-  #       format.html do
-  #         redirect_to invoices_path, notice: t('.already_paid') and return
-  #       end
-
-  #       format.json { render json: @payment_order.errors, status: :unprocessable_entity and return }
-  #     end
-  #   end
-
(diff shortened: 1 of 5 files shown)


maricavor (approved):

vohmar:
outdated with new vat rate change

PR #1244, opened 2024-04-18
added secure ssl condition 2

OlegPhenomenon:
**List of changes:**

1. Removed the "no content" issue that appeared where the notification bell is located.
2. Updated the websockets that send notifications (the bell should light up red).
3. Fixed the issue where clicking on a notification causes the notification window to appear.
4. Changed the UI of the notification window.
5. When pressing the notification button, the bell should change from red to white, indicating that the notifications have been read.
6. Fixed error messages and successful operation messages for the wishlist page.
7. Added domain validation when a user attempts to enter a domain name to add to the wishlist.
8. Changed the loading principle of Google Analytics, thereby eliminating some warnings in devtool.
9. Eliminated the error warning of loading rails-ujs.
10. Changed the dropdown button for notifications.
11. Fixed a webpush notification error since the browser security policy was updated and the old solution did not work.
12. Changed issues with logging out - it was making a GET request instead of DELETE.
13. Wherever there was a link for deleting anything (offer, profile, etc.) on the auction site, I changed it to a button because pressing DELETE was making a GET request. This is related to the peculiarities of Turbo's operation.
14. Fixed "some number cannot be parsed" issues; there was a situation where a web socket would send a form where one should enter a bid, but instead of displaying the minimum available bid, it sent an empty form with no values.
15. Fixed system tests.

**What would I test?**

1. The wishlist page, all the logic for adding, changing, and adding correct/correct domain names.
2. All types of notifications: web-push, displaying notifications as a bell, and as a label. Also, this needs to be done in the case of websockets.
3. Try deleting bids, billing profiles, users, also on the admin side.
4. Make bids from two browsers and check if valid data and a valid form for adding the bid arrive.
5. Check the grammar and meaning of error messages and successful actions.


vohmar:
https://auction.infra.tld.ee/cookies?cookies=declined
internal error, when declining all cookies

web push - pressed agree, popup went away, but the browser notification dialog did not appear. Popup went away because the permission to show notification on broser side was already there. Once i changed that the webpush agree button did not do anyting any more. Not oce was the bowser dialog called out in chrome nor MS edge

webpush available button was active but nothing happened with pressing it
after refreshing the page the webpush available button became inactive

cleared cookies, set browser notifications to ask, logged out and logged back in again web push notification appeared as expected, but the agree button did not do anything this time (same in edge and chrome)

in the notification list (bell icon) new notification are listed at the end of the list - perhaps its better to reverse the order and have the newest on top of the list as user cannot delist the messages.

Issue #1275, opened 2024-06-27
Make a deposit view has three decimal places and in min bid info and '.' as decimal separator


vohmar:
![image](https://github.com/internetee/auction_center/assets/7734747/a1146e87-ad8b-41aa-bde6-257353c0e3e5)

* dot as decimal separator - minimum bid, deposit required, explanation text. Should be always comma
* three decimal places in minimum bid. Should be two decimal places

PR #1277, opened 2024-06-28
Comma issue

OlegPhenomenon:
fixes https://github.com/internetee/auction_center/issues/1275


vohmar:
now it's dot in estonian and english. Please use comma as a decimal separator regardless of the language setting

image

vohmar:
Can we just set comma as a decimal separator regardless of language?

dot used for english for deposits (back button translation missing):
image

comma used in text box:
image

PR #1284, opened 2024-07-03
fix for show action in admin controller of auction

### LINKED ISSUE #1282: Internal error on accessing running auction's detail view in admin
https://auction.internet.ee/admin/auctions

clicking on any name in the list will result in internal error - 404 (Not Found)

OlegPhenomenon:
close #1282


vohmar:
the problem is fixed for the auctions that are set up, but the issue is still there for those that still do not have start and end time set 

image

PR #1285, opened 2024-07-03
decrease space between confirmation and send code buttons

### LINKED ISSUE #1279: Too wide gap between phone confirmation submit and send button
image

due to the big gap that does not look good, the code sending button is hidden way down in the mobile view and people do not notice it.

OlegPhenomenon:
close #1279


vohmar:
cannot make any bids not pay the deposit with the user that needs to confirm their contacts. If i press the bid or pay the deposit button. In mobile view i am seeing the content missing message. In console I see 500 (Internal Server Error).

image

